PR #2, opened 2020-02-10
initial iam operator commit

posriniv:


posriniv:
Will handle build problems later on

PR #19, opened 2020-02-25
Modify to allow namespaced operator to deploy cluster role

yannizhang2019:


yannizhang2019:
This change allows namespaced operator to deploy cluster role and cluster role binding according to https://github.com/operator-framework/operator-sdk/blob/master/doc/operator-scope.md#cluster-scoped-operator-usage
With this change, cluster role ibm-iam-operator is created on the cluster

PR #55, opened 2020-03-24
Change management ingress container port to 443

liqlin2015:
Because of catalog UI redirect problem it has to change container
port of management ingress from 8443 to 443.


posriniv:
@liqlin2015 there is no change in the service right? Only the container port has changed, isn't it?

liqlin2015:
Yes @posriniv . Have verified in Steve's cluster. Metering now is working fine.

PR #59, opened 2020-03-26
change image and channel version

posriniv:


deploy/olm-catalog/ibm-iam-operator/3.5.0/ibm-iam-operator.v3.5.0.clusterserviceversion.yaml
```diff
@@ -585,7 +585,7 @@ spec:
   maintainers:
   - email: support@ibm.com
     name: IBM Support
-  maturity: alpha
+  maturity: stable-v1
```
DanielXLee: Cannot use `stable-v1` in here, OLM doesn't support this value, you can use `stable`

posriniv: okay

PR #65, opened 2020-04-16
fix s390x base image

posriniv:


raajsekhar:
Thanks @posriniv, .. 
when can we expect the image published into https://quay.io/repository/opencloudio/ibm-iam-operator .. so that i can continue the CS-3.3 Verification on s390x

posriniv:
@raajsekhar it should be pushed by the cicd build. Can you please try pulling the latest image and check?

PR #82, opened 2020-04-28
add icp_port env variable.

morvencao:
This PR tries to add env variable:
```
ICP_PORT
```
The env variable is useful for OIDC provider to expose public endpoints to external OIDC client.


morvencao:
@yannizhang2019 @posriniv PTAL

PR #123, opened 2020-06-22
enable nonce by default

rashmi43:


rashmi43:
@posriniv Do we need to update older CSV's? I think for ifix2 we will have 3.6.4 version and that will have nonce value as true since the authentications_cr has the value as true.

rashmi43:
It will be to 3.6.4 because for ifix1, we created 3.6.3 and there will be a new csv version for 3.6.4. Removed the change to 3.6.3

PR #146, opened 2020-08-03
add clusterrole and clusterrolebinding to support cloudpakswitchers

posriniv:


pkg/controller/authentication/clusterrolebinding.go
```diff
@@ -96,6 +96,19 @@ func generateCRBData(defaultAdminUser string, oidcIssuerURL string) map[string]C
 			},
 			RoleName: "cluster-admin",
 		},
+		"cloudpak-switchers-binding": CRBData{
+			Subject: []SubjectData{
+				{
+					Name: oidcIssuerURL + "#" + defaultAdminUser,
```
yannizhang2019, line 102: what is this user for?

pkg/controller/authentication/clusterrolebinding.go
```diff
@@ -96,6 +96,19 @@ func generateCRBData(defaultAdminUser string, oidcIssuerURL string) map[string]C
...
+					Kind: "User",
+				},
+				{
+					Name: defaultAdminUser,
```
yannizhang2019, line 106: only default admin has this role?

PR #148, opened 2020-08-11
Update configmap to set roks parameters

yannizhang2019:
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/39426

### DIFF
--- a/pkg/controller/authentication/configmap.go
+++ b/pkg/controller/authentication/configmap.go
@@ -79,42 +79,6 @@ func (r *ReconcileAuthentication) handleConfigMap(instance *operatorv1alpha1.Aut
 		return nil
 	}
 
-	/*
-		wellknownURL := "https://kubernetes.default:443/.well-known/oauth-authorization-server"
-		tokenFile := "/var/run/secrets/kubernetes.io/serviceaccount/token"
-		caCert, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/ca.crt")
-		if err != nil {
-			reqLogger.Error(err, "Failed to read ca cert", "ConfigMap.Namespace", instance.Namespace, "ConfigMap.Name")
-		}
-		caCertPool := x509.NewCertPool()
-		caCertPool.AppendCertsFromPEM(caCert)
-		content, err := ioutil.ReadFile(tokenFile)
-		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/controller/authentication/configmap.go
```diff
@@ -45,40 +52,37 @@ func (r *ReconcileAuthentication) handleConfigMap(instance *operatorv1alpha1.Aut
 	err = r.client.Get(context.TODO(), types.NamespacedName{Name: consoleConfigMapName, Namespace: instance.Namespace}, consoleConfigMap)
 	if err != nil {
 		if errors.IsNotFound(err) {
-			reqLogger.Error(err, "The configmap ", consoleConfigMapName ," is not created yet")
-			return err
-		} else {
-			reqLogger.Error(err, "Failed to get ConfigMap",  consoleConfigMapName)
+			reqLogger.Error(err, "The configmap ", consoleConfigMapName, " is not created yet")
```
posriniv, line 55: @yannizhang2019 why did you remove the else block here

pkg/controller/authentication/configmap.go
```diff
@@ -45,40 +52,37 @@ func (r *ReconcileAuthentication) handleConfigMap(instance *operatorv1alpha1.Aut
...
 	proxyConfigMapName := "ibmcloud-cluster-info"
 	proxyConfigMap := &corev1.ConfigMap{}
 	err = r.client.Get(context.TODO(), types.NamespacedName{Name: proxyConfigMapName, Namespace: instance.Namespace}, proxyConfigMap)
 	if err != nil {
 		if errors.IsNotFound(err) {
-			reqLogger.Error(err, "The configmap ", proxyConfigMapName ," is not created yet")
-			return err
-		} else {
-			reqLogger.Error(err, "Failed to get ConfigMap",  proxyConfigMapName)
+			reqLogger.Error(err, "The configmap ", proxyConfigMapName, " is not created yet")
```
posriniv, line 69: @yannizhang2019 why did you remove the else block here?

pkg/controller/authentication/configmap.go
yannizhang2019: that's from go lint

pkg/controller/authentication/configmap.go
yannizhang2019: go lint flagged it, so I changed based on go lint recommendation, we still have a lot go lint errors

PR #154, opened 2020-08-20
Make scopes configurable in Liberty

rashmi43:


pkg/apis/operator/v1alpha1/authentication_types.go
```diff
@@ -108,6 +108,9 @@ type ConfigSpec struct {
 	ROKSURL                     string `json:"roksURL"`
 	ROKSUserPrefix              string `json:"roksUserPrefix"`
 	EnableImpersonation         bool   `json:"enableImpersonation"`
+	CLAIMS_SUPPORTED            string `json:"claimsSupported"`
```
posriniv, line 111: Please follow the coding convention here, don't use all caps

posriniv (commented):
You will have to edit the csv file as well

pnpavan25 (approved):
lgtm

deploy/crds/operator.ibm.com_authentications_crd.yaml
```diff
@@ -274,6 +285,7 @@ spec:
               - imageName
               - imageRegistry
               - imageTag
+              - scopeClaim
```
znhafeez, line 288: @rashmi43 does `scopeClaim` need to be under the `required` field? If so, we need to add it in the CR for `identity-provider`.

deploy/crds/operator.ibm.com_authentications_crd.yaml
```diff
@@ -240,6 +248,7 @@ spec:
               required:
               - imageName
               - imageRegistry
+              - bootstrapUserId
```
znhafeez, line 251: Is `boostrapUserId` a required field or a config field? We can either remove it from this required field list or, if it is required, we can add it to the CR and place it under the imageName, imageRegistry, etc. fields.

deploy/crds/operator.ibm.com_authentications_crd.yaml
znhafeez: I'm not sure of the details of the changes but I think it should be removed from here because it's part of the config

deploy/crds/operator.ibm.com_v1alpha1_authentication_cr.yaml
```diff
@@ -96,6 +96,10 @@ spec:
     openshiftPort: 443
     roksEnabled: true
     nonceEnabled: true
+    claimsSupported: "name,family_name,display_name,given_name,preferred_username"
+    claimsMap: "name=\"givenName\" family_name=\"givenName\" given_name=\"givenName\" preferred_username=\"givenName\"
+    display_name=\"displayName\""
+    scopeClaim: "profile=\"name,family_name,display_name,given_name,preferred_username\""
     bootstrapUserId: ""
```
znhafeez, line 103: bootstraUserId should be on like 51 if it's a required field

deploy/crds/operator.ibm.com_authentications_crd.yaml
rashmi43: i have removed it from required list

deploy/crds/operator.ibm.com_v1alpha1_authentication_cr.yaml
rashmi43: bootStrap is what i expect the user to set, for the other three there are defaults. But should be editable if customer wants.

pnpavan25 (approved):
lgtm

PR #169, opened 2020-08-28
Issue40633 - Resolve upgrade issue, make non required

rashmi43:
Removed fields as required as they cause issues with the upgrade scenario.


posriniv (approved):
lgtm

PR #174, opened 2020-09-01
Updated latest image sha digest for platform-auth and identity-provider container images for 3.7.1

pnpavan25:
Updated latest image sha digest for platform-auth and identity-provider container images for 3.7.1


posriniv (approved):
lgtm

PR #176, opened 2020-09-01
Updated new image sha digest for platform-auth, identity-provider and identity-manager container images.

pnpavan25:
Updated new image sha digest for platform-auth, identity-provider and identity-manager container images.


posriniv (approved):
lgtm

PR #177, opened 2020-09-02
Updated latest image sha digest for IAM operand images.

pnpavan25:
Updated latest image sha digest for IAM operand images.


posriniv (approved):
lgtm

yannizhang2019:
Can we keep a note on where the sha values are from?

PR #197, opened 2020-09-16
Updated latest image sha digest for IAM operand images.

pnpavan25:
Updated latest image sha digest for IAM operand images.


posriniv (approved):
lgtm

PR #199, opened 2020-09-17
Updated latest sha digest for IAM operand identity-manager image.

pnpavan25:
Updated latest sha digest for IAM operand identity-manager image.


rashmi43 (approved):
lgtm, I will update the certification image.

PR #229, opened 2020-10-30
Auditlogging upgrade

znhafeez:
Issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/41447

Addressing upgrade scenario for new audit syslog sidecar.

### DIFF
--- a/pkg/controller/policydecision/policydecision_controller.go
+++ b/pkg/controller/policydecision/policydecision_controller.go
@@ -513,6 +513,11 @@ func (r *ReconcilePolicyDecision) ingressForPolicyDecision(instance *operatorv1a
 
 func (r *ReconcilePolicyDecision) deploymentForPolicyDecision(instance *operatorv1alpha1.PolicyDecision) *appsv1.Deployment {
 
+	// Update the audit image for upgrade scenarios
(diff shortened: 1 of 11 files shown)


pkg/controller/policydecision/policydecision_controller.go
```diff
@@ -513,6 +513,11 @@ func (r *ReconcilePolicyDecision) ingressForPolicyDecision(instance *operatorv1a
 
 func (r *ReconcilePolicyDecision) deploymentForPolicyDecision(instance *operatorv1alpha1.PolicyDecision) *appsv1.Deployment {
 
+	// Update the audit image for upgrade scenarios
+	if instance.Spec.AuditService.ImageName != "audit-syslog-service" {
+		instance.Spec.AuditService.ImageName = "audit-syslog-service"
```
yannizhang2019, line 518: audit-syslog-service is being referenced in multiple places, should we create a constant variable to hold this value?

znhafeez: added a constants.go file where we can hold variables like this

yannizhang2019 (approved):

PR #287, opened 2021-01-20
changes to support v3 channel

posriniv:


horis233:
@posriniv It looks good to me.

PR #405, opened 2021-08-18
Automate next csv and update to 3.11.2, update operand versions.

rashmi43:
Update to 3.11.2 and update operand versions


smagesh2106 (approved):
igtm, make sure to test it.

PR #442, opened 2021-11-09
Update ibm-iam-operator.v3.12.1.clusterserviceversion.yaml  for IAM uses image tags in relatedImages, instead of digests issue

duaribiju:


rashmi43 (commented):
need to make for 3.12.2, that PR is not yet committed

PR #488, opened 2022-08-10
Add endpoint init

ash007-ibm:
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/53620

Change in brief -> Add init container to check two crucial endpoints before starting the actual job. The script to check endpoints is available in iam-onboarding job.


pkg/controller/securityonboarding/securityonboarding_controller.go
```diff
@@ -355,7 +361,8 @@ func (r *ReconcileSecurityOnboarding) handleJob(instance *operatorv1alpha1.Secur
 	foundErr2 := false
 	iamJobExists := false
 	if err != nil {
-		reqLogger.Info("Failed to create iam-onboarding Job, exists already ", "Job.Namespace", instance.Namespace, "Job.Name", "iam-onboarding")
+		reqLogger.Info("The err from IAMOnboardJob is ")
```
ujjwalchk-it, line 364: Can we keep existing log template - which is standard?

pkg/controller/securityonboarding/securityonboarding_controller.go
```diff
@@ -368,6 +375,7 @@ func (r *ReconcileSecurityOnboarding) handleJob(instance *operatorv1alpha1.Secur
 			reqLogger.Info("Successfully created Job", "Job.Namespace", instance.Namespace, "Job.Name", "iam-onboarding")
 		}
 	} else if restartRequired {
+		reqLogger.Info("Restart required for iamOnboardJob")
```
ujjwalchk-it, line 378: Same , keep existing logging template.

pkg/controller/securityonboarding/securityonboarding_controller.go
```diff
@@ -765,6 +923,45 @@ func getIAMOnboardJob(instance *operatorv1alpha1.SecurityOnboarding, r *Reconcil
 					corev1.ResourceMemory: *memory128},
 			},
 		},
+		{
+			Name:            "endpoints-test",
```
ujjwalchk-it, line 927: Can we rename this to -> "init-token-validation"

pkg/controller/securityonboarding/securityonboarding_controller.go
```diff
@@ -1116,6 +1163,68 @@ func getIAMOnboardJob(instance *operatorv1alpha1.SecurityOnboarding, r *Reconcil
 	currentJob := &batchv1.Job{}
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: "iam-onboarding", Namespace: instance.Namespace}, currentJob)
 	if err == nil {
+		if currentJob.Status.Conditions != nil && currentJob.Status.Conditions[0].Type == "Failed" {
```
ujjwalchk-it, line 1166: I think we need to add proper exception handling here to make sure the code is not broken.
eg.
```
config, err := rest.InClusterConfig()
  if err != nil {
		  return config, err
	  }
```

ujjwalchk-it (changes requested):
Few negative testcases for the log collection -

1. What would be happen if the log regex failed? `regexp.Compile(":\\s?(.*)")`
2. What would happen if matching logic failed? `				matches := r.FindAllString(events.Items[len(events.Items)-3].Message, -1)
`
3. I would suggest as we have multiple checks there , we should test with various combinations before we take this.

ash007-ibm:
> 

   1. What would be happen if the log regex failed? regexp.Compile(":\\s?(.*)")
    Compile function will fail if the regex expression (`:\\s?(.*)`) is invalid, as given expression is valid this would never throw error.
  2.  What would happen if matching logic failed? matches := r.FindAllString(events.Items[len(events.Items)-3].Message, -1)
I have handled that scenario with else in place (https://github.com/IBM/ibm-iam-operator/pull/488/files#diff-5496a5426f771b114da6a7b071729d240cf7f25a87db2d1d817a45b41ea810edR1222)
  3.   I would suggest as we have multiple checks there , we should test with various combinations before we take this.
We have in only 3 combination in  iam-onboarding job's event in given order

-   Created pod: iam-onboarding--1-xxxx (5 times)

-  Delete pod: iam-onboarding--1-xxxx (5 times)

-  Job has reached the specified backoff limit 

And once the job fails, last event is `Job has reached the specified backoff limit`.
In code I am consuming created pod event, if for whatever reason it has consumed the last event. I have else check in place mentioned above.

ash007-ibm:
Final PR has been raised in  -https://github.com/IBM/ibm-iam-operator/pull/492

PR #503, opened 2022-10-05
Remove policy controller, and cluster role related code

yannizhang2019:


yannizhang2019:
cannot resolve conflicts

PR #539, opened 2023-01-09
Begin phasing out use of ROKS_ENABLED for OS Auth

rwhundley:
- Check value of `OSAUTH_ENABLED` in the platform-auth-idp ConfigMap instead of `ROKS_ENABLED` for creating and deleting OAuthClients
- Add logging around the values of these variables when considering creation of OAuthClients


rashmi43:
hi @rwhundley today when I installed the latest build, I realised if we rely only on OSAUTH the oauthclient for Zen won't be created, which is fine but tomorrow when I enable it Zen login will fail because the previously created zenclient will not have a corresponding oauthclient. 

Let's discuss this with @yannizhang2019 and @Tirumalavasa please ask Yanni if it's okay to have oauthclient create permission in the IAM operator csv. You can start a discussion thread on slack with Yanni.

rashmi43 (approved):

PR #570, opened 2023-02-07
 [CP3.0] - Update authentication_controller.go to fix calls to v1alpha certs

rashmi43:
do not watch v1alpha certificates on CP3


sgrube (commented):
/lgtm

PR #574, opened 2023-02-09
Update routes

rwhundley:
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57040

remove mount of namespace-scope configmap and remove extra routes

### DIFF
--- a/pkg/controller/authentication/containers.go
+++ b/pkg/controller/authentication/containers.go
@@ -984,17 +984,6 @@ func buildIdentityManagerContainer(instance *operatorv1alpha1.Authentication, id
 			Name:  "MASTER_HOST",
 			Value: icpConsoleURL,
 		},
-		{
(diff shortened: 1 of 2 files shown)


pkg/controller/authentication/containers.go
```diff
@@ -984,17 +984,6 @@ func buildIdentityManagerContainer(instance *operatorv1alpha1.Authentication, id
 			Name:  "MASTER_HOST",
 			Value: icpConsoleURL,
 		},
-		{
-			Name: "NAMESPACES",
-			ValueFrom: &corev1.EnvVarSource{
```
rashmi43, line 989: there are two references to NAMESPACES in identity-manager but they are not used in the CP3 context, I will test with the merge build and confirm that nothing breaks

rashmi43 (approved):

PR #594, opened 2023-03-03
Derive shared services namespace from Authentication

rwhundley:
Addresses the following:

- https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57489
- https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57490

Changes:

- Perform an intelligent guess on where the shared services namespace is based upon where the Operator is able to list an Authentication CR. There should only ever be one present among the namespaces the Operator can list. If no or more than one Authentication CR is found, an error is thrown as this is an unsupported usage of the Authentication CR.
- Use the shared services namespace for lookups of ConfigMaps, Secrets, etc.
- Revert use of namespace-to-ClientControllerConfig map to just a single ClientControllerConfig on ReconcileClient because the resources being looked up should always be in the shared services namespace, thus no longer requiring a map structure.
- When handling URLs that the code knows are based upon Service names, always convert them to the format `<Service name>.<Namespace of the Service>.svc`.

### DIFF
--- a/pkg/apis/addtoscheme_oidc_v1.go
+++ b/pkg/apis/addtoscheme_oidc_v1.go
@@ -19,6 +19,7 @@ package apis
 import (
 	v1 "github.com/IBM/ibm-iam-operator/pkg/apis/oidc/v1"
   oauthv1 "github.com/openshift/api/oauth/v1"
+  v1alpha1 "github.com/IBM/ibm-iam-operator/pkg/apis/operator/v1alpha1"
 )
 
 func init() {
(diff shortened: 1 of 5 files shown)


rashmi43 (commented):
these changes look good Rob but I was wondering is there any impact to the authIdpConfigMap creation in SOD, the platform-auth-idp configmap has these variables:
```
"IDENTITY_AUTH_DIRECTORY_URL":        "https://platform-auth-service:3100",
			"IDENTITY_PROVIDER_URL":              "https://platform-identity-provider:4300",
			"IDENTITY_MGMT_URL":                  "https://platform-identity-management:4500",

PR #597, opened 2023-03-08
[DNM]adjust probes

Tirumalavasa:
The intention of this adjustment is that we wanted to catch if platform-auth-service is down(may be restarting) and  restart platform-identity-provider pod when platform-auth-service restarts so that boot logic present in the provider can send saml/oidc/ldap xmls back to auth-service as they would be lost due to auth-service pod restart.


Tirumalavasa:
closing this PR, not required

PR #598, opened 2023-03-08
Fix .status.service naming collisions

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#57621](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57621)

The first pass at implementing .status.service block for ODLM tried to use the Kubernetes object name as a unique id for entries in the list of managed resources. This assumption fails because names are only unique among their GVK, not necessarily elsewhere. As a result, Service and Deployment name collisions occurred, which led to new Deployment-related state being written to the indices where Services of the same name were stored, leaving the old Deployment state in the latter indices never to be touched again.

This fix replaces all of the extra fuss around what the previous state of the object was and instead replaces .status.service with the .status.service that was most recently obtained from the current cluster state and performs an update only if they do not deeply equal one another.

### DIFF
--- a/pkg/apis/operator/v1alpha1/authentication_types.go
+++ b/pkg/apis/operator/v1alpha1/authentication_types.go
@@ -18,6 +18,7 @@ package v1alpha1
 
 import (
   "context"
+  "reflect"
   "sync"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


Daniel-Fan (approved):
/lgtm

PR #603, opened 2023-03-10
Fix operator reference in CSV

ashank07:
In latest CSV update, operator image was updated to point at our operand location (`icr.io/cpopen/cpfs` instead of the operator location `icr.io/cpopen`).


ashank07:
Force merging this as I think we may have already lost Rob for the day and we want to get this into a build so we can get testing done over the weekend.  Let me know if this is a problem.

PR #607, opened 2023-03-16
Service Upgrade handling for IM4.0 

ujjwalchk-it:
Handle update for both `label` and `podselector` during the CP3 upgrade for the below services :
```
platform-auth-service,
platform-identity-management,
platform-identity-provider
```
Operator logs -
```
{"level":"info","ts":1678972184.4709521,"logger":"controller_authentication","msg":"Upgrade check : Successfully updated service podSelector , label details ","Instance.Namespace":"ibm-common-services","Instance.Name":"example-authentication","Service.Name":"platform-auth-service"}
{"level":"info","ts":1678972184.4709523,"logger":"controller_authentication","msg":"Upgrade check : Successfully updated service podSelector , label details ","Instance.Namespace":"ibm-common-services","Instance.Name":"example-authentication","Service.Name":"platform-identity-provider"}
{"level":"info","ts":1678972184.563555,"logger":"controller_authentication","msg":"Upgrade check : Successfully updated service podSelector , label details ","Instance.Namespace":"ibm-common-services","Instance.Name":"example-authentication","Service.Name":"platform-identity-management"}
```

Screenshot 2023-03-16 at 6 48 14 PM

### DIFF
--- a/pkg/controller/authentication/service.go
+++ b/pkg/controller/authentication/service.go
@@ -46,6 +46,33 @@ func (r *ReconcileAuthentication) handleService(instance *operatorv1alpha1.Authe
 	} else if err != nil {
 		reqLogger.Error(err, "Failed to get Service")
 		return err
+	} else {
+		// Going to validate pod-selector for CP3 upgrade
+		var cp2podselector bool = false
+		var cp2podlabel bool = false
+		podSelector := currentService.Spec.Selector
+		value, ok := podSelector["k8s-app"]
+		if ok && value != "platform-auth-service" {
+			currentService.Spec.Selector = map[string]string{"k8s-app": "platform-auth-service"}
+			cp2podselector = true
+		}
+		// Going to validate label for CP3 upgrade
+		label := currentService.Labels
+		value, ok = label["app"]
+		if ok && value != "platform-auth-service" {
+			currentService.Labels = map[string]string{"app": "platform-auth-service"}
+			cp2podlabel = true
+		}
+		if cp2podselector || cp2podlabel {
+			err = r.client.Update(context.Background(), currentService)
+			if err != nil {
+				reqLogger.Error(err, "Upgrade check : Failed to update service podSelector , label details ", "Service.Namespace", instance.Namespace, "Service.Name", "platform-auth-service", "Error.message", err)
+				r.needToRequeue = true
+				return err
+			} else {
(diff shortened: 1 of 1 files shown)


rashmi43 (approved):

pkg/controller/authentication/service.go
```diff
@@ -46,6 +46,33 @@ func (r *ReconcileAuthentication) handleService(instance *operatorv1alpha1.Authe
 	} else if err != nil {
 		reqLogger.Error(err, "Failed to get Service")
 		return err
+	} else {
```
rwhundley, line 49: It looks as if the following code could/should be broken out into its own function and called once for each of these services. Something like this maybe?

```golang
// validateCP3ServicePodSelectorAndLabel takes a *Service and attempts to update that Service's selectors and
// labels if they do not match the desired CP3 values. Returns an error if the update fails or if the provided
// *Service is nil or lacks a name or namespace.
func validateCP3PodSelectorAndLabel(currentService *corev1.Service) (err error) {
	if currentService == nil || currentService.Name == "" || currentService.Namespace == "" {
		return fmt.Errorf("received invalid Service")
	}
	var cp2podselector bool = false
	var cp2podlabel bool = false
	podSelector := currentService.Spec.Selector
	value, ok := podSelector["k8s-app"]
	if ok && value != currentService.Name {
		currentService.Spec.Selector = map[string]string{"k8s-app": currentService.Name}
		cp2podselector = true
	}
	// Going to validate label for CP3 upgrade
	label := currentService.Labels
	value, ok = label["app"]
	if ok && value != currentService.Name {
		currentService.Labels = map[string]string{"app": currentService.Name}
		cp2podlabel = true
	}
	if cp2podselector || cp2podlabel {
		err = r.client.Update(context.Background(), currentService)
		if err != nil {
			reqLogger.Error(err, "Upgrade check : Failed to update service podSelector , label details ", "Service.Namespace", instance.Namespace, "Service.Name", currentService.Name, "Error.message", err)
			r.needToRequeue = true
			return
		} else {
			reqLogger.Info("Upgrade check : Successfully updated service podSelector , label details ", "Service.Name", currentService.Name)
		}
	}
	return
}

ujjwalchk-it:
GH Ref https://github.ibm.com/IBMPrivateCloud/roadmap/issues/57787

pkg/controller/authentication/service.go
ujjwalchk-it: Thanks for your suggestion , I changed accordingly.

rashmi43 (approved):

PR #624, opened 2023-04-20
Remove dependency on Routes API

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#58183](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58183)

Now, the Operator will only build with and handle Schemes related to OpenShift if the Operator detects that it has been deployed on an OpenShift cluster. Routes in this case specifically are skipped if running on a non-OpenShift cluster.

Other Changes:

* Created function for getting the shared services namespace from a CommonService CR, which is a more straightforward approach than trying to list Authentications.
* Reorganized some utils-like files into common directories to clean up organization of pkg code.


rashmi43 (commented):
generateRandomstring is missing

pkg/controller/authentication/configmap.go
```diff
@@ -316,14 +314,13 @@ func (r *ReconcileAuthentication) handleConfigMap(instance *operatorv1alpha1.Aut
 				}
 
 				_, keyExists := currentConfigMap.Data["IS_OPENSHIFT_ENV"]
-				//currentConfigMap.Data["IS_OPENSHIFT_ENV"] = strconv.FormatBool(isOSEnv)
 				if keyExists {
 					reqLogger.Info("Current configmap", "Current Value", currentConfigMap.Data["IS_OPENSHIFT_ENV"])
-					if currentConfigMap.Data["IS_OPENSHIFT_ENV"] != strconv.FormatBool(isOSEnv) {
-						currentConfigMap.Data["IS_OPENSHIFT_ENV"] = strconv.FormatBool(isOSEnv)
+					if currentConfigMap.Data["IS_OPENSHIFT_ENV"] != strconv.FormatBool(r.RunningOnOpenShiftCluster()) {
```
rashmi43, line 319: @Tirumalavasa do you see this causing any issues. we kind of always assume it to be true. Not sure how it will behave on CNCF, we had another check in operands as well to see if its CNCF

rashmi43 (approved):

PR #625, opened 2023-04-24
Issue 46862

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#46862](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/46862)


rwhundley:
Not going to go this way for this problem; that plus other Operators have entrenched themselves further in this kind of leftover resource problem, so the urgency on solving this here is down.

PR #639, opened 2023-04-30
Update client_controller.go to enhance secret cache

rashmi43:


rwhundley:
I think we should instead do what I've laid out in #640 - this caching logic isn't really needed because the controller runtime already has its own cache that it maintains. No need for us to define a different one in our own code.

PR #651, opened 2023-05-26
Fix ownership of ibmcloud-* resources issues

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#58841](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58841)

During backup and restore, the ibmcloud-cluster-info ConfigMap and the ibmcloud-cluster-ca-cert Secret would block the IM Operator from coming completely online because it expects that it owns those resources; however, when the resources were restored, the ownerReferences were not also restored. This fix to the IM Operator attempts to assert controller status on behalf of the Authentication CR upon the ibmcloud-cluster-info ConfigMap when it isn't controlled by that Authentication CR, and, in the case of ibmcloud-cluster-ca-cert, the IM Operator will simply delete the Secret and create a new one to replace it.

### DIFF
--- a/pkg/controller/authentication/configmap.go
+++ b/pkg/controller/authentication/configmap.go
@@ -392,7 +392,7 @@ func (r *ReconcileAuthentication) handleConfigMap(instance *operatorv1alpha1.Aut
 					err = fmt.Errorf("an error occurred during registration-json generation")
 					return err
 				}
-				reqLogger.Info("Calculated new platform-oidc-registration.json", "json", newConfigMap.Data["platform-oidc-registration.json"])
+				reqLogger.Info("Calculated new platform-oidc-registration.json")
 				var currentRegistrationJSON, newRegistrationJSON *registrationJSONData
 				newRegistrationJSON = &registrationJSONData{}
 				currentRegistrationJSON = &registrationJSONData{}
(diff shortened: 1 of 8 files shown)


pkg/controller/authentication/configmap.go
```diff
@@ -100,22 +101,29 @@ func (r *ReconcileAuthentication) handleConfigMap(instance *operatorv1alpha1.Aut
...
 				*needToRequeue = true
 				return
 			} else {
-				reqLogger.Info("Can't determine the configmap ownership , IM installation may not proceed further until the configmap is removed", "Configmap.Namespace", currentConfigMap.Namespace, "ConfigMap.Name", "ibmcloud-cluster-info")
+				reqLogger.Info("ConfigMap is not owned by the current Authentication or a ManagementIngress; will attempt to become controller", "Configmap.Namespace", currentConfigMap.Namespace, "ConfigMap.Name", "ibmcloud-cluster-info")
+				if err = controllerutil.SetControllerReference(instance, currentConfigMap, r.client.Scheme()); err != nil {
+					reqLogger.Error(err, "Could not become the controller of the ConfigMap; it may need to be removed", "Configmap.Namespace", currentConfigMap.Namespace, "ConfigMap.Name", "ibmcloud-cluster-info")
+					return
+				}
+				currentConfigMap.Labels["app"] = "auth-idp"
```
rashmi43, line 119: Is this label correct?

rwhundley: Based upon what installs on the cluster today in v3, yes.

rashmi43, line 119: which configmap is this?

rashmi43 (approved):

PR #652, opened 2023-05-29
keep route certificate details

rashmi43:
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58063

### DIFF
--- a/pkg/controller/authentication/routes.go
+++ b/pkg/controller/authentication/routes.go
@@ -328,6 +328,22 @@ func (r *ReconcileAuthentication) reconcileRoute(ctx context.Context, instance *
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


rwhundley (changes requested):

pkg/controller/authentication/routes.go
```diff
@@ -328,6 +328,22 @@ func (r *ReconcileAuthentication) reconcileRoute(ctx context.Context, instance *
 			}
 		}
 
+		reqLogger.Info("Found key in observed route", observedRoute.Spec.TLS.Key)
```
rwhundley, line 331: We shouldn't log out credentials like this by default; if we want to leave these logs in, please use `reqLogger.V(<some number above 0).Info()` instead so that they only show up at higher logging levels.

rwhundley (changes requested):

PR #659, opened 2023-06-22
fix for dynamic scan

vijayanjay:
Fixes - https://github.ibm.com/IBMPrivateCloud/roadmap/issues/59023


rashmi43 (approved):
ingresses are not used in cd but we can keep this change

PR #704, opened 2023-10-17
labels addition throguh CR

Tirumalavasa:
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/60601

### DIFF
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -174,14 +174,15 @@ func GetSharedServicesNamespace(ctx context.Context, name string) (namespace str
 }
 
 func MergeMap(in map[string]string, mergeMap map[string]string) map[string]string {
+	//in map is always Not NIL -- from the function calls, we ensure in map always has some data in it
 	if in == nil {
 		in = make(map[string]string)
 	}
 	if mergeMap == nil {
 		mergeMap = make(map[string]string)
 	}
-	for k, v := range mergeMap {
-		in[k] = v
+	for k, v := range in {
+		mergeMap[k] = v
 	}
-	return in
+	return mergeMap
 }
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -172,3 +172,16 @@ func GetSharedServicesNamespace(ctx context.Context, name string) (namespace str
 	}
 	return
 }
+
+func MergeMap(in map[string]string, mergeMap map[string]string) map[string]string {
+	if in == nil {
+		in = make(map[string]string)
+	}
+	if mergeMap == nil {
+		mergeMap = make(map[string]string)
+	}
+	for k, v := range mergeMap {
+		in[k] = v
+	}
+	return in
+}
--- a/pkg/common/utils.go
+++ b/pkg/common/utils.go
@@ -174,12 +174,11 @@ func GetSharedServicesNamespace(ctx context.Context, name string) (namespace str
 }
 
 func MergeMap(in map[string]string, mergeMap map[string]string) map[string]string {
-	//in map is always Not NIL -- from the function calls, we ensure in map always has some data in it
-	if in == nil {
-		in = make(map[string]string)
-	}
 	if mergeMap == nil {
 		mergeMap = make(map[string]string)
+	}
+	if in == nil {
+		return mergeMap
 	}
 	for k, v := range in {
 		mergeMap[k] = v
--- a/pkg/controller/authentication/deployment.go
+++ b/pkg/controller/authentication/deployment.go
@@ -22,6 +22,7 @@ import (
 	gorun "runtime"
 
 	operatorv1alpha1 "github.com/IBM/ibm-iam-operator/pkg/apis/operator/v1alpha1"
+	pkgCommon "github.com/IBM/ibm-iam-operator/pkg/common"
 	"github.com/IBM/ibm-iam-operator/pkg/controller/shatag"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -305,11 +306,21 @@ func generateDeploymentObject(instance *operatorv1alpha1.Authentication, scheme 
 	ldapCACert := instance.Spec.AuthService.LdapsCACert
 	routerCertSecret := instance.Spec.AuthService.RouterCertSecret
 
+	metaLabels := pkgCommon.MergeMap(map[string]string{"app": deployment}, instance.Spec.Labels)
+	podMetadataLabels := map[string]string{
+		"app":                        deployment,
+		"k8s-app":                    deployment,
+		"component":                  deployment,
+		"app.kubernetes.io/instance": "platform-auth-service",
+		"intent":                     "projected",
+	}
+	podLabels := pkgCommon.MergeMap(podMetadataLabels, instance.Spec.Labels)
+
 	idpDeployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      deployment,
 			Namespace: instance.Namespace,
-			Labels:    map[string]string{"app": deployment},
+			Labels:    metaLabels,
 		},
 		Spec: appsv1.DeploymentSpec{
 			Replicas: &replicas,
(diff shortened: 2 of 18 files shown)


ujjwalchk-it (approved):

rashmi43 (commented):
can you list all the resources that will have labels @Tirumalavasa

pkg/common/utils.go
```diff
@@ -172,3 +172,16 @@ func GetSharedServicesNamespace(ctx context.Context, name string) (namespace str
 	}
 	return
 }
+
+func MergeMap(in map[string]string, mergeMap map[string]string) map[string]string {
+	if in == nil {
+		in = make(map[string]string)
+	}
+	if mergeMap == nil {
+		mergeMap = make(map[string]string)
```
rwhundley, line 181: Shouldn't we just return in the event that `mergeMap == nil`? It's essentially a no-op on `in`, which we know at this point is at least an empty `map[string]string`.

pkg/controller/authentication/deployment.go
```diff
@@ -305,11 +306,21 @@ func generateDeploymentObject(instance *operatorv1alpha1.Authentication, scheme
...
 
+	metaLabels := pkgCommon.MergeMap(map[string]string{"app": deployment}, instance.Spec.Labels)
+	podMetadataLabels := map[string]string{
+		"app":                        deployment,
+		"k8s-app":                    deployment,
+		"component":                  deployment,
+		"app.kubernetes.io/instance": "platform-auth-service",
+		"intent":                     "projected",
+	}
+	podLabels := pkgCommon.MergeMap(podMetadataLabels, instance.Spec.Labels)
```
rwhundley, line 317: If the intent is for the `podMetadataLabels` to be preserved, then the arguments should be reordered.

pkg/controller/authentication/deployment.go
```diff
@@ -305,11 +306,21 @@ func generateDeploymentObject(instance *operatorv1alpha1.Authentication, scheme
 	ldapCACert := instance.Spec.AuthService.LdapsCACert
 	routerCertSecret := instance.Spec.AuthService.RouterCertSecret
 
+	metaLabels := pkgCommon.MergeMap(map[string]string{"app": deployment}, instance.Spec.Labels)
```
rwhundley, line 309: If the intent is for the `app` label value to be preserved, then the arguments should be reordered.

pkg/controller/authentication/job.go
```diff
@@ -91,6 +92,13 @@ func generateJobObject(instance *operatorv1alpha1.Authentication, scheme *runtim
 	reqLogger := log.WithValues("Instance.Namespace", instance.Namespace, "Instance.Name", instance.Name)
 	image := shatag.GetImageRef("IM_INITCONTAINER_IMAGE")
 	resources := instance.Spec.ClientRegistration.Resources
+
+	metaLabels := pkgCommon.MergeMap(map[string]string{"app": jobName}, instance.Spec.Labels)
```
rwhundley, line 96: If the intent is for the `app` label value to be preserved, then the arguments should be reordered.

pkg/controller/authentication/job.go
```diff
@@ -91,6 +92,13 @@ func generateJobObject(instance *operatorv1alpha1.Authentication, scheme *runtim
...
+	podMetaLabels := map[string]string{
+		"app":                        jobName,
+		"app.kubernetes.io/instance": "oidc-client-registration",
+	}
+	podLabels := pkgCommon.MergeMap(podMetaLabels, instance.Spec.Labels)
```
rwhundley, line 101: If the intent is for the `podMetaLabels` to be preserved, then the arguments should be reordered.

pkg/controller/authentication/routes.go
```diff
@@ -458,6 +459,9 @@ func (r *ReconcileAuthentication) newRoute(instance *operatorv1alpha1.Authentica
 
 	weight := int32(100)
 
+	commonLabel := map[string]string{"app": "im"}
+	routeLabels := pkgCommon.MergeMap(commonLabel, instance.Spec.Labels)
```
rwhundley, line 463: If the intent is for the `commonLabel` labels to be preserved, then the arguments should be reordered.

rwhundley (changes requested):

pkg/common/utils.go
Tirumalavasa: hi @rwhundley , Nil check for `in` map is not really required, as from the MergeMap function calls, i ensured `in` map always has data, I just added nil check for `in` map too thinking in future if someone wants to consume MergeMap function we validate `in` is not nil.

pkg/controller/authentication/deployment.go
Tirumalavasa: Hi @rwhundley , instead of swapping the arguments in this function, In the MergeMap function body i have swapped it, meaning that i have considered mergeMap as a source(instance.Spec.Labels) to add back podMetadataLables to mergeMap, so that we preserve the podMetadataLabels

pkg/controller/authentication/deployment.go
Tirumalavasa: instead of swapping the arguments in this function, In the MergeMap function body i have swapped it, meaning that i have considered mergeMap as a source(instance.Spec.Labels) to add back podMetadataLables to mergeMap, so that we preserve the podMetadataLabels

pkg/controller/authentication/deployment.go
```diff
@@ -436,11 +441,21 @@ func generateProviderDeploymentObject(instance *operatorv1alpha1.Authentication,
...
```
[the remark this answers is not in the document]
Tirumalavasa: instead of swapping the arguments in this function, In the MergeMap function body i have swapped it, meaning that i have considered mergeMap as a source(instance.Spec.Labels) to add back app label to mergeMap, so that we preserve app label

pkg/controller/authentication/deployment.go
```diff
@@ -567,11 +576,21 @@ func generateManagerDeploymentObject(instance *operatorv1alpha1.Authentication,
...
+	podMetadataLabels := map[string]string{
+		"app":                        deployment,
+		"k8s-app":                    deployment,
+		"component":                  deployment,
+		"app.kubernetes.io/instance": "platform-identity-management",
+		"intent":                     "projected",
+	}
+	podLabels := pkgCommon.MergeMap(podMetadataLabels, instance.Spec.Labels)
```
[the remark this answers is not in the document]
Tirumalavasa: instead of swapping the arguments in this function, In the MergeMap function body i have swapped it, meaning that i have considered mergeMap as a source(instance.Spec.Labels) to add back podMetadataLables to mergeMap, so that we preserve the podMetadataLabels

pkg/controller/authentication/job.go
Tirumalavasa: instead of swapping the arguments in this function, In the MergeMap function body i have swapped it, meaning that i have considered mergeMap as a source(instance.Spec.Labels) to add back app label to mergeMap, so that we preserve app label

pkg/controller/authentication/routes.go
Tirumalavasa: i have  swapped it in the source function(MergeMap) body

pkg/common/utils.go
```diff
@@ -174,14 +174,15 @@ func GetSharedServicesNamespace(ctx context.Context, name string) (namespace str
 }
 
 func MergeMap(in map[string]string, mergeMap map[string]string) map[string]string {
+	//in map is always Not NIL -- from the function calls, we ensure in map always has some data in it
```
rwhundley, line 177: I would rather remove this line - this is already defensively coded such that the `nil` case is handled (either var is replaced with empty `map[string]string` in that event), but this comment conflicts with the logic.

pkg/common/utils.go
rwhundley: Right, but we should be coding defensively here against misuse. What we don't want are situations where binaries panic. That said, the way this is written now will be fine given we are still making empty `map[string]string` for both vars. I was just saying that, if one is non-`nil` and the other is `nil`, we could just return whichever value has been submitted (a no-op.)

rwhundley (changes requested):

PR #711, opened 2023-11-22
Issue 60940

rwhundley:
[Originating issue](https://github.ibm.com/ibmprivatecloud/roadmap/issues/60940)

Reconcile logic was missing returns for errors found during creates or
updates of Client CRs, which led to failing CRs never reconciling
properly unless retriggered with a small update.

Other changes:

* Remove vestigial CACHE_UPDATE check

### DIFF
--- a/pkg/controller/client/client_controller.go
+++ b/pkg/controller/client/client_controller.go
@@ -174,10 +174,6 @@ func (r *ReconcileClient) Reconcile(ctx context.Context, request reconcile.Reque
 	} else {
 		reqLogger.Info("successfully set config for controller")
 	}
-	// If this is a cache update only, return
-	if request.Name == "CACHE_UPDATE" {
-		return reconcile.Result{}, nil
-	}
 
 	reqLogger.Info("Reconciling OIDC Client data")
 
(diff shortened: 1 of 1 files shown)


pkg/controller/client/client_controller.go
```diff
@@ -333,14 +329,22 @@ func (r *ReconcileClient) processOidcRegistration(ctx context.Context, client *o
...
 		err = r.createClient(ctx, client)
 	} else {
 		reqLogger.Info("Client found, update Client")
+		verb = "update"
 		err = r.updateClient(ctx, client)
 	}
 
+	if err != nil {
+		reqLogger.Error(err, fmt.Sprintf("Error occured while attempting to %s Client", verb))
+		return
```
rashmi43, line 345: should we return the error as well?

rashmi43 (approved):

pkg/controller/client/client_controller.go
rwhundley: This is a naked return from a function with named returns, which means that `return` is the same as `return err` in this context.

rwhundley: Example of how this is used here: [A Tour of Go](https://go.dev/tour/basics/7)

rashmi43 (approved):

pkg/controller/client/client_controller.go
rashmi43: ok

PR #715, opened 2023-12-01
Upgrade operator-sdk to v1 stream

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#58284](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58284)

This commit largely focuses on upgrading the Operator from the v0.19.x
operator-sdk to v1.x. This represents a significant update to the
internals as far as dependencies go, but the exposed behavior should not
have changed in a way that breaks compatibility.

Some significant refactorings are included as a part of this upgrade; they
are as follows:

* Use [subreconciler helpers](https://github.com/opdev/subreconciler/tree/main)
  to compose reconciliation logic in the Client controller.
* Remove client handling functions that had become passthroughs.
* Rename ClientControllerConfig to AuthenticationConfig; the
  configuration was never describing the Client controller, but instead
  the Authentication CR that the Client controller is interacting with
  on the cluster.
* Remove the AuthenticationConfig from the ClientReconciler type
  definition and instead construct the *AuthenticationConfig as a step
  in each of the subreconcilers and pass it to functions as an argument;
  this reduces concerns about potentially stale state being accessed on
  the ClientReconciler.
* Revert from using the CommonService CR to the Authentication CR
  search-based method for determining the services namespace to avoid
  dependency issues.
* Add various interfaces and structs for errors produced in the Client
  controller runtime.
* Rename some files and move around some function - icperror.go became
  errors.go, the Zen registration-specific function was moved to
  zen_registration.go to mirror clientreg.go for OIDC registration.
* Replace the custom type for the conditions reported in Clients and the
  functions built around that type with the k8s library equivalents to
  reduce the amount of custom code we need to maintain.
* Remove redundant HTTP method constants already provided by net/http package.
* Add the `cs-ca-certificate-secret` Secret to the AuthenticationConfig
  to allow for more configuration to be passed into functions instead of
  calling a separate k8s Client.Get call.
* Add Makefile targets from operator-sdk v1 to support new
  layout/format.


rashmi43:
I am good with these changes. what about the unit tests @rwhundley ?

Tirumalavasa:
I am good with the changes

rashmi43 (approved):

PR #744, opened 2024-02-08
[INPROGRESS]consume postgresql configmap and secret deployed by odlm

Tirumalavasa:
https://github.ibm.com/ibmprivatecloud/roadmap/issues/62210

### DIFF
--- a/controllers/operator/operandrequest.go
+++ b/controllers/operator/operandrequest.go
@@ -18,11 +18,12 @@ package operator
 
 import (
 	"context"
-	"fmt"
 
 	operatorv1alpha1 "github.com/IBM/ibm-iam-operator/apis/operator/v1alpha1"
+	"github.com/IBM/ibm-iam-operator/controllers/common"
 	ctrlCommon "github.com/IBM/ibm-iam-operator/controllers/common"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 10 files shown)


controllers/operator/operandrequest.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+		reqLogger.Info("Failed to get OperandRequest " + ctrlCommon.IMEDBOprName)
+		return false
+	}
+	reqLogger.Info("OperandRequest " + ctrlCommon.IMEDBOprName + " is present")
+
+	return true
+}
+
+func (r *AuthenticationReconciler) checkCommonServiceExists(instance *operatorv1alpha1.Authentication) (exist bool) {
```
rwhundley, line 182: Given that the code isn't doing anything more than trying to create this CommonService CR if it doesn't already exist (versus actually performing comparisons to see if there is a change that needs to be made by the reconciler), breaking out a separate "check" function can be skipped - the logic should just try to create the resource and, if the resource already exists, api will give us back that error, which we can identify to know that there's no more work to do.

rwhundley (changes requested):

rashmi43 (approved):

Tirumalavasa:
Closing this PR, as this is taken care as part of https://github.com/IBM/ibm-iam-operator/pull/745

PR #751, opened 2024-02-21
Disable DDL execution until Operands are updated

rwhundley:
Providing this PR as a means of backing out EDB schema initialization from the Operands if the images are not updated in time for the next build.


rashmi43 (approved):

rashmi43:
operand changes will be ready only by next week

PR #760, opened 2024-03-06
Add delays to Authentication requeues

rwhundley:
More requeues happening in the Operator's controllers leads to an increased propensity to hit exponential backoff. Given the Authentication controller only manages a single CR instance, the reconciliation logic can afford to provide more explicit, quick turnarounds for loops without having to worry too much about the impact on the overall rate limit.


Tirumalavasa (approved):
/lgtm

PR #765, opened 2024-03-08
zen_instances_users table column uz_id changed to uid

Tirumalavasa:


controllers/operator/migration/migration.go
```diff
@@ -766,15 +766,15 @@ func insertZenInstanceUsers(ctx context.Context, mongodb *MongoDB, postgres *Pos
...
 			"UserID":        zenInstanceUser.UserID,
 			"ZenInstanceID": zenInstanceUser.ZenInstanceID,
 		}
 
 		query := `
 			INSERT INTO platformdb.zen_instances_users
-			(uz_id, zen_instance_id, user_id)
-			VALUES (@UZID, @ZenInstanceID, @UserID)
+			(uid, zen_instance_id, user_id)
+			VALUES (@UID, @ZenInstanceID, @UserID)
```
rwhundley, line 777: This will not work - `_id` in ZenInstanceUsers is a MongoDB ObjectId, but the DDL has changed this to be a UUID. To do this, you'll need to swap `@UID` for `DEFAULT`.

rwhundley (changes requested):

controllers/operator/migration/migration.go
Tirumalavasa: @rwhundley , Thank you for pointing this, I just correct it.
replaced `@UID` with `DEFAULT`

rwhundley (approved):

PR #767, opened 2024-03-12
idp_configs: change ldap_config(json) to ldap_id(string)

Tirumalavasa:
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/62461


controllers/operator/migration/schema/v1/tables.go
```diff
@@ -198,6 +198,17 @@ func ConvertToIdpConfig(idpMap map[string]interface{}, idpConfig *IdpConfig) (er
 	} else {
 		idpMap["enabled"] = true
 	}
+	// SAML with LDAP dependency mongo document will have ldap_config: {ldap_id: <value>}
+	// which directly maps to ldap_id column in sql
+	if ldap_config, ok := idpMap["ldap_config"]; ok {
+		if ldap_config, ok := ldap_config.(map[string]interface{}); ok {
+			// If ldap_config is already a map, fetch the ldap_id
+			ldap_id := ldap_config["ldap_id"]
```
rwhundley, line 206: I used same `val, ok` pattern for this assignment as the previous ones you used to avoid setting `ldap_id` in this case to `nil`, which is what would happen if this key ever wasn't set on `ldap_config`.

controllers/operator/migration/schema/v1/tables.go
```diff
@@ -198,6 +198,17 @@ func ConvertToIdpConfig(idpMap map[string]interface{}, idpConfig *IdpConfig) (er
...
+			delete(idpMap, "ldap_config") // delete the ldap_config json and introduce ldap_id string
```
rwhundley, line 207: This deletion isn't necessary - we've removed the json struct tag for `ldap_config` and replaced it with the `ldap_id` struct tag, so `ldap_config` isn't used again after the unmarshal happens.

rwhundley (commented):
I've noted the relevant differences between your implementation and mine; let me know what you want to do (if anything.)

Tirumalavasa:
> I've noted the relevant differences between your implementation and mine; let me know what you want to do (if anything.)

Nothing much Rob, i just tried following your approach for uniformity

controllers/operator/migration/schema/v1/tables.go
rwhundley: I agree with this: 
> it is the responsibility of the operands to read and if this has proper value or not
All I am saying is that the same end is reached by skipping inclusion of the `ldap_id` key entirely vs. setting it to an empty value. That said, I don't think it's worth holding up an approval over.

PR #784, opened 2024-03-26
Update Non Zen User roles for isolated upgrade

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#62770](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/62770)

* Query Team collection in MongoDB to determine initial role value for User rows

### DIFF
--- a/migration/migration.go
+++ b/migration/migration.go
@@ -682,6 +682,12 @@ func insertUserRelatedRows(ctx context.Context, mongodb *MongoDB, postgres *Post
 			err = nil
 			continue
 		}
+		if _, err = setUserV3Role(ctx, mongodb, user); err != nil {
+			reqLogger.Error(err, "Failed to set role for user")
+			errCount++
+			err = nil
+			continue
+		}
 		var uid *uuid.UUID
 		if uid, err = insertUser(ctx, postgres, user); err != nil {
 			errCount++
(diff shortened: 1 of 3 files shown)


migration/migration.go
```diff
@@ -843,6 +849,42 @@ func setClientSecretIfEmpty(ctx context.Context, mongodb *MongoDB, zenInstance *
...
+		Find(ctx, filter)
+
+	var highest v1schema.Role
+	for cursor.Next(ctx) {
+		team := &v1schema.Team{}
+		if err = cursor.Decode(&team); err != nil {
+			return
+		}
+
+		if current := getHighestRoleForUserFromTeam(user.UserID, team); current > highest {
```
rashmi43, line 879: we need to assess group membership to determine role as well: https://github.ibm.com/IBMPrivateCloud/platform-identity-provider/blob/4.4.0/server/boot/z_cp3migration.js#L453

rashmi43: looks like the above is not used. A user may belong to multiple teams.

migration/schema/v1/tables.go
```diff
@@ -1011,3 +1011,113 @@ func ConvertV2SamlToIdpConfig(samlMap map[string]any) (v3Config *IdpConfig, err
...
+		return Authenticated
+	}
+}
+
+// ToV3String returns an IM-compatible role name as a string. In IM, there are only three valid roles - Administrator,
+// Viewer, and Authenticated.
+func (r Role) ToV3String() (s string) {
+	if r >= Administrator && r <= ClusterAdmin {
+		return Administrator.ToString()
+	} else if r >= Viewer && r < Administrator {
```
rashmi43, line 1078: There are two roles Administrator role and empty role
const CP3_AUTHENTICATED_ROLE = '';

rashmi43 (approved):
looks okay, what if user role is via group membership?

PR #787, opened 2024-03-27
SCIM JIT Migration [Not ready]

abhishekp03:


migration/schema/v1/tables.go
```diff
@@ -1011,3 +1011,39 @@ func ConvertV2SamlToIdpConfig(samlMap map[string]any) (v3Config *IdpConfig, err
...
+// Group is respretation of a row in "platformdb"."groups" table
+type Group struct {
+	GroupID     string `json:"group_id,omitempty"`
+	DisplayName string `json:"display_name,omitempty"`
+	RealmID     string `json:"realm_id,omitempty"`
+}
+
+func ConvertToGroup(grp map[string]any) (g *Group) {
+	group := &Group{}
+	group.GroupID = grp["_id"].(string)
```
rwhundley, line 1024: Given we're operating on `any`/`interface{}` types, trying to blindly assert a type can lead to a panic. For example, if `grp` was an empty `map[string]any`, `grp["_id"]` is `nil`, and asserting that `nil` is a `string` will cause the program to panic, e.g. [example](https://go.dev/play/p/9qnMNVQp9hf). This is why the two-value test form is used for accessing map values with keys and asserting types on interfaces. Examples of this can be found in the ConvertV2DirectoryToV3IdpConfig function.

migration/schema/v1/tables.go
```diff
@@ -1011,3 +1011,39 @@ func ConvertV2SamlToIdpConfig(samlMap map[string]any) (v3Config *IdpConfig, err
 	v3Config.Enabled = true
 	return
 }
+
+// Group is respretation of a row in "platformdb"."groups" table
+type Group struct {
+	GroupID     string `json:"group_id,omitempty"`
```
rwhundley, line 1017: As a convention, I'm avoiding `omitempty` for any columns that have `NOT NULL` on them, so I'd prefer to see these removed.

migration/schema/v1/tables.go
```diff
@@ -1011,3 +1011,39 @@ func ConvertV2SamlToIdpConfig(samlMap map[string]any) (v3Config *IdpConfig, err
...
+func ConvertToGroup(grp map[string]any) (g *Group) {
+	group := &Group{}
+	group.GroupID = grp["_id"].(string)
+	group.DisplayName = grp["displayName"].(string)
+	group.RealmID = "defaultSP"
+	return group
+}
+
+// UserGroup is respretation of a row in "platformdb"."users_groups" table
+type UserGroup struct {
```
rwhundley, line 1031: I assume we're still waiting on an insert UserGroup implementation, correct?

rwhundley (changes requested):

migration/schema/v1/tables.go
abhishekp03: That's correct.

migration/migration.go
```diff
@@ -1138,4 +1139,167 @@ func removeInvalidUserPreferences(usersPrefs []v1schema.UserPreferences) (update
...
 }
 
+func insertGroups(ctx context.Context, mongodb *MongoDB, postgres *PostgresDB) (err error) {
+	dbName := "platform-db"
+	collectionName := "Groups"
+	reqLogger := logf.FromContext(ctx).WithValues("MongoDB.DB", dbName, "MongoDB.Collection", collectionName)
+	filter := bson.M{
+		"$and": bson.A{
+			bson.M{"migrated": bson.M{"$ne": true}},
+			bson.M{"type": bson.M{"$regex": "//^SAML$//i"}},
```
rwhundley, line 1149: So one of two suggestions here; either:
- Just do a direct string comparison - I don't know that the case sensitivity handling should be important, and regexes can be annoying to deal with
- If we need the regex, then this needs to be fixed as this will not work. It would be easier to guarantee that this does what you want with the following form:
```suggestion
			bson.M{"type": bson.M{"$regex": "^SAML$", "$options": "i"}},
```
Alternatively, I think there may be a type in the primitives package that is used for defining regexes.

rwhundley (changes requested):

abhishekp03:
Closing this PR in favour of the updated PR https://github.com/IBM/ibm-iam-operator/pull/797

PR #792, opened 2024-03-29
Add MigrationsPerformed condition

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#62506](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/62506)

* Add the MigrationsPerformed condition to the Authentication CR, which communicates whether all necessary migrations have been performed. This status factors into the `.status.service` that the Authentication reports - when one or more migrations are active or have failed, the Authentication is flagged as NotReady.


Tirumalavasa (approved):
/lgtm

PR #794, opened 2024-03-29
Update Go 1.21.7 -> 1.21.8

rwhundley:
Additionally

* Fix `fmt` target so that it no longer uses a hard-coded version value for the Go sdk path.


Tirumalavasa (approved):
/lgtm

PR #804, opened 2024-04-12
Make migration keys more unique

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#62995](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/62995)

Previously, in multi-Pak migration where two Paks migrate from one common install of Common Services to two separate installs, one Pak's migration activity written back to MongoDB was indistinguishable from another's, which led to a race condition where the first Pak to migrate would be successful and the second would fail due to skipping all of the documents that had been migrated by the first Pak.

This change uses the services namespace as a part of the key written to the migrated MongoDB documents to avoid this naming collision between migrations.


Tirumalavasa:
Cloing this PR in favor of https://github.com/IBM/ibm-iam-operator/pull/830

PR #820, opened 2024-05-07
Issue-63379: Fix for nil pointer error if user has already been migrated

abhishekp03:
Summary of changes:
1. Fixing bug where nil was being returned if a user already existed in the users table in EDB, causing nil pointer errors in further flows.


Tirumalavasa (approved):
/lgtm

PR #828, opened 2024-05-14
[CD] Update zen-instance for client CR update

ujjwalchk-it:
Issue => https://github.ibm.com/IBMPrivateCloud/roadmap/issues/52227

Manual test 
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/52227#issuecomment-79473729

Regression & Manual CPD test with custom catalog
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/52227#issuecomment-79473922


controllers/oidc.security/client_controller.go
```diff
@@ -496,6 +489,22 @@ func (r *ClientReconciler) processZenRegistration(ctx context.Context, req ctrl.
...
 	}
 
+	//Zen registration exist, update
+	if zenReg != nil {
+		//Zen registration exists - going to update zen instance registration
+		reqLogger.Info("Zen registration already exists for oidc client - the zen instance will be updated",
+			"clientId", clientCR.Spec.ClientId, "zenInstanceId", clientCR.Spec.ZenInstanceId)
+		if err = r.registerZenInstance(ctx, clientCR, clientCreds, config); err != nil {
+			reqLogger.Error(err, "Failed to update Zen registration")
+			if err := r.writeErrorConditionsAndEvents(ctx, clientCR, err, http.MethodPost); err != nil {
```
rashmi43, line 499: should we use err2 here

rashmi43 (approved):

controllers/oidc.security/client_controller.go
ujjwalchk-it: I think we should be good given for the initial error (err) we are calling the reconciler with error `return subreconciler.RequeueWithError(err)` , so the next check won't even happen. It is the same way being implemented in the existing flow.

Tirumalavasa (approved):
/lgtm

PR #840, opened 2024-06-08
Update to version 4.7.0

DaniyalMustafa:


Tirumalavasa:
@DaniyalMustafa , Common services version is 4.8.0, we should just be bumping it to 4.7.0 and IM operator 4.7 will get shipped as part of CS 4.8

Tirumalavasa (approved):
/lgtm

PR #848, opened 2024-07-08
Remove redundant endpoint - idauth

YangShyrMei:
The redundant endpoint "idauth" used to originally redirect to itself. Thus, seeing an increase in number of security vulnerabilities and observe the need to remove it.
Parent issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/60011

### DIFF
--- a/controllers/operator/ingress.go
+++ b/controllers/operator/ingress.go
@@ -97,7 +97,6 @@ func (r *AuthenticationReconciler) handleIngress(instance *operatorv1alpha1.Auth
 		idmgmtV2ApiIngress,
 		platformAuthIngress,
 		platformIdAuthBlockIngress,
-		platformIdAuthIngress,
 		platformIdProviderIngress,
 		platformLoginIngress,
 		platformOidcBlockIngress,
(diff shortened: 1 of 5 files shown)


Tirumalavasa (approved):

controllers/operator/ingress.go
```diff
@@ -372,7 +371,7 @@ func platformIdAuthBlockIngress(instance *operatorv1alpha1.Authentication, schem
 						HTTP: &netv1.HTTPIngressRuleValue{
 							Paths: []netv1.HTTPIngressPath{
 								{
-									Path:     "/idauth/oidc/endpoint",
+									Path:     "/oidc/endpoint",
```
rashmi43, line 374: we can remove this as well

rwhundley (changes requested):
Requesting changes because I think this may be missing a chunk of the logic related to what @rashmi43 mentioned re: tokens and want to be sure that's not needed before this gets merged. It seems like the OIDC registration.json needs to be updated to use the new value set in controllers/operators/constants.go, and that JSON needs to be applied by re-running the Job that submits it, but I don't see anything related to that in here. I've also just added some testing around the Route reconciliation in the Authentication controller, so I'd imagine those will need to be updated as well.

rwhundley:
Ah, my mistake re: testing - missed that this was going into SC2. The comment re: test updates does not apply here, but the other parts of my review do.

YangShyrMei:
Will build a new pr with the different approach as used in https://github.com/IBM/ibm-iam-operator/pull/871

PR #866, opened 2024-08-22
Remove Routes when Zen front door is enabled

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#63956](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/63956)

When the `zenFrontDoor` config value is set to `true`, the IM Routes
are removed from the cluster. Otherwise, the default of creating and
updating the Routes happens.

Additionally:

* Refactor Route handling to subreconcilers
* Add testing for Route handling
* Tweak Route handling logs


rashmi43 (approved):

controllers/operator/routes.go
```diff
@@ -285,7 +287,7 @@ func (r *AuthenticationReconciler) reconcileRoute(ctx context.Context, instance
 			// Requeue after creating new route
 			*needToRequeue = true
 		}
-	} else {
+	} else if shouldHaveRoutes(instance) {
```
sgrube, line 290: Should this be checked before the route get above is done?  If the flag is on, the get above will throw an error and shouldn't.

controllers/operator/routes.go
```diff
@@ -247,6 +247,10 @@ func (r *AuthenticationReconciler) handleRoutes(ctx context.Context, instance *o
 	return
 }
 
+func shouldHaveRoutes(instance *operatorv1alpha1.Authentication) bool {
+	return !instance.Spec.Config.ZenFrontDoor
```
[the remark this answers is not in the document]
rwhundley: If the flag doesn't exist, it will default to its zero value, `false`, which is the default for the field. This seems fine to me, but let me know if you disagree.

controllers/operator/routes.go
rwhundley: That ends up not being an issue, but something _later on_ does. Fixing that.

controllers/operator/routes.go
```diff
@@ -328,17 +330,26 @@ func (r *AuthenticationReconciler) reconcileRoute(ctx context.Context, instance
...
+			err = r.Update(ctx, observedRoute)
 			if err != nil {
 				reqLogger.Error(err, "Failed to update route")
 				return
 			}
 			*needToRequeue = true
 		}
+	} else {
+		reqLogger.Info("zenFrontDoor is enabled; deleting route")
+		err = r.Delete(ctx, observedRoute)
```
rwhundley, line 346: This is an issue - if the Route doesn't exist, observedRoute is empty, meaning we hit a client error trying to delete something without naming it first.

controllers/operator/routes.go
rwhundley: Also going to clean this up a little bit given it took multiple passes for the issue to be seen.

PR #887, opened 2024-09-25
Skip migrating broken Zen documents

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#64896](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64896)

When ZenInstance documents are broken, e.g. the OIDC clients referred to by them no longer exist, they break migration. Whenever this happens, instead skip these documents as well as ZenInstanceUser documents that refer to the skipped ZenInstance.

Additionally, introduce a subreconciler to the Client controller to wait for the Authentication CR to become ready before attempting to manage OIDC client registrations.


ujjwalchk-it (commented):
lgtm

PR #888, opened 2024-09-25
Ensure MigrationsPerformed is True when done migrating

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#64955](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64955)

When attempting to override whether the migration should be considered complete or not, e.g. setting `"authentication.operator.ibm.com/migration-complete"` to `"true"` when some migration failure has resulted in some non-essential data never reaching EDB, it could get the Authentication stuck in an InProgress state.

This adds a subreconciler that ensures that, upon exiting the migration subreconciler, the Authentication CR always has the MigrationsPerformed condition set to True.

### DIFF
--- a/controllers/operator/authentication_controller.go
+++ b/controllers/operator/authentication_controller.go
@@ -45,6 +45,7 @@ import (
 
 	operatorv1alpha1 "github.com/IBM/ibm-iam-operator/apis/operator/v1alpha1"
 	"github.com/opdev/subreconciler"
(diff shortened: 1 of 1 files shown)


Tirumalavasa (approved):
/lgtm

Tirumalavasa:
@rwhundley , We need to back port this change to `master` (CD stream)as well right ?

PR #912, opened 2024-11-13
Reconcile Routes and ZenExtension later

rwhundley:
Originating issue: [PrivateCloud-analytics/Zen#39543](https://github.ibm.com/PrivateCloud-analytics/Zen/issues/39543)

If Zen and IM were installed on the cluster at the same time, and the Zen front door was enabled on the Authentication CR, this would lead to IM waiting for the ZenExtension resource to be reconciled without standing up its Operands for Zen to register OIDC clients with.

This changes the order in which objects are reconciled so that Zen will be able to communicate with IM's Operands and create those OIDC clients, which will in turn allow the Zen Operator to handle the ZenExtensions that IM needs in order to expose its Operands to ingress traffic.


Tirumalavasa (approved):
/lgtm

PR #925, opened 2025-01-06
Always update values in platform-auth-idp when triggered

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#65604](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65604)

* Add a check to be sure that, if a selected key is not present in the observed ConfigMap that is present in the generated one, the selected key is added to the observed ConfigMap.
* Always update DEFAULT_LOGIN to whatever is set in the Authentication CR.
* Update test to account for this change.


Tirumalavasa (approved):
/lgtm

PR #926, opened 2025-01-08
Updated golang.org/x/net v0.33.0 & golang.org/x/crypto v0.31.0[SC2]

Raniprathyusha:


go.mod
```diff
@@ -93,6 +93,6 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-replace golang.org/x/crypto => golang.org/x/crypto v0.17.0
+replace golang.org/x/crypto => golang.org/x/crypto v0.31.0
```
rwhundley, line 96: You can remove both `replace` lines because the indirects already use the correct version.

rwhundley (changes requested):

PR #948, opened 2025-02-05
Rework ibmcloud-cluster-info reconcile

rwhundley:
The `ibmcloud-cluster-info` ConfigMap is now reconciled as follows:

* If the `cluster_address_auth` field is not present in either case (e.g. upgrade from IM version without this field), it is set to match the `cluster_address` field's value.
* If the Zen front door is disabled, the ConfigMap is initially created with the default cp-console values and is left unmanaged after that point.
* If the Zen front door is enabled, all hostname values for `cluster_address`, `cluster_address_auth`, `cluster_endpoint`, and `proxy_address` are aligned to match the hostname provided by Zen.


Tirumalavasa (approved):
/lgtm

PR #958, opened 2025-02-12
[SC2] Skip MongoDB bootstrap CM update when not present

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#65879](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65879)

### DIFF
--- a/controllers/operator/migration.go
+++ b/controllers/operator/migration.go
@@ -209,11 +209,11 @@ func (r *AuthenticationReconciler) handleMongoDBCleanup(ctx context.Context, req
 			return subreconciler.RequeueWithError(err)
 		} else if objUpdated {
 			anyObjUpdated = true
-			reqLogger.Info("Object was finalized",
+			reqLogger.V(1).Info("Object was finalized",
 				"Object.Name", kvp.key.Name,
 				"Object.Namespace", kvp.key.Namespace)
 		} else {
-			reqLogger.Info("Object was not found",
+			reqLogger.V(1).Info("Object was not found",
 				"Object.Name", kvp.key.Name,
 				"Object.Namespace", kvp.key.Namespace)
 		}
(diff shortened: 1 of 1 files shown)


controllers/operator/migration.go
```diff
@@ -547,6 +547,20 @@ func (r *AuthenticationReconciler) overrideMongoDBBootstrap(ctx context.Context,
...
 	}
 
+	// find icp-mongodb-init
+	icpMongoDBInitCMKey := types.NamespacedName{Namespace: req.Namespace, Name: "icp-mongodb-init"}
+	icpMongoDBInitCM := &corev1.ConfigMap{}
+	reqLogger.V(1).Info("Get MongoDB bootstrap ConfigMap",
+		"ConfigMap.Name", icpMongoDBInitCMKey.Name,
+		"ConfigMap.Namespace", icpMongoDBInitCMKey.Namespace)
+	if err = r.Get(ctx, icpMongoDBInitCMKey, icpMongoDBInitCM); k8sErrors.IsNotFound(err) {
+		reqLogger.Info("No MongoDB bootstrap ConfigMap to patch; continuing")
```
Tirumalavasa, line 557: Hi @rwhundley , This code changes look fine, But i am just thinking other approach where we can just call the `overrideMongoDBBootstrap` only if `hasPreloadMongoDBConfigMap` returns `false` which indicates it is in-place upgrade. just to avoid a GET calls to Configmap and Statefulset here

rwhundley: The GET calls should be hitting the cache if they're truly unnecessary, so I'm not too worried about that. My thought is that the preload ConfigMap should serve as a clear signal that some MongoDB migration activity is required, not that it has to be an isolated upgrade. Given we're trying to make a quick fix, I'd err on the side of avoiding too much additional churn as there's enough cognitive load from having to consider each of the different potential install cases already.

rwhundley: Also, there should really never be a time where the icp-mongodb-init ConfigMap is there but the icp-mongodb StatefulSet isn't - I just wanted to be extra certain that we skipped trying to go through this whole process of updating internal certificates if we didn't have 100% of the required objects present for that to work.

Tirumalavasa: ok

Tirumalavasa (approved):

ujjwalchk-it (approved):

PR #959, opened 2025-02-12
define registry under cpfs

rashmi43:
Gaps from olm-utils discussion


rwhundley:
We've gotten this addressed in some of our other PRs; closing.

PR #965, opened 2025-02-21
AllNamespace upgrade- check mongo deployment in instance namespace

Tirumalavasa:
Fixes:
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65916


ujjwalchk-it (approved):
As discussed , unlike inplace upgrade we are expecting the mongo operator/operands to be running in the separate ns where as im-op runs on the common openshift-operator ns. So in addition to check in the pod_namespace(openshift-operator) , we should also check for the instance instance (authentication CR ns).

change looks good me .  This shouldn't impact for isolate scenario but SERT needs to validate inplace & allnamespace upgrade.

PR #971, opened 2025-02-25
Rename charts and templates

rwhundley:
* Renames chart name fields to comply with IBM formats
* Renames cluster-scoped templates to enforce a render order


helm-cluster-scoped/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
-name: ibm-iam-operator-cluster-scoped-resources
+name: ibm-iam-operator-cluster-scoped
 description: Cluster-scoped Resources for IBM IAM Operator
 type: application
-version: "4.11.0"
+version: "4.11.0+20250225.133605.1"
```
ashank07, line 5: Please remove the build metadata.  This will be added by the build pipeline and should not be included in the operator repo.

helm/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: ibm-iam-operator
 description: IBM IAM Operator
 type: application
-version: "4.11.0"
+version: "4.11.0+20250225.133605.1"
```
ashank07, line 5: same here, revert the build metadata

ashank07 (changes requested):

rashmi43 (approved):

PR #977, opened 2025-03-10
passthrough route for cert auth

Tirumalavasa:
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65599


controllers/operator/routes.go
```diff
@@ -555,6 +566,18 @@ func (r *AuthenticationReconciler) newRoute(authCR *operatorv1alpha1.Authenticat
 			InsecureEdgeTerminationPolicy: routev1.InsecureEdgeTerminationPolicyRedirect,
 			DestinationCACertificate:      string(fields.DestinationCAcert),
 		}
+	} else if fields.RoutePath == "" {
+		// Passthrough route (if RoutePath is empty)
+		route.Spec.TLS = &routev1.TLSConfig{
+			Termination:                   routev1.TLSTerminationPassthrough,
+			InsecureEdgeTerminationPolicy: routev1.InsecureEdgeTerminationPolicyRedirect,
+		}
+	} else {
```
ujjwalchk-it, line 575: As discussed we can remove it for now as it is not used.

Tirumalavasa: @ujjwalchk-it , removed the edge termination block

controllers/operator/routes.go
```diff
@@ -321,22 +331,23 @@ func (r *AuthenticationReconciler) reconcileRoute(authCR *operatorv1alpha1.Authe
...
-		return subreconciler.ContinueReconciling()
-	} else if err != nil {
-		reqLogger.Error(err, "Failed to get existing route for reconciliation")
-		return subreconciler.RequeueWithError(err)
-	}
-	err = r.Delete(ctx, observedRoute)
-	if err != nil {
-		reqLogger.Error(err, "Failed to delete the Route")
-		return subreconciler.RequeueWithError(err)
+	if fields.Name != IMCrtAuthRouteName {
```
rwhundley, line 334: I'd prefer inverting this check to `fields.Name == IMCrtAuthRouteName` and immediately returning with `subreconciler.ContinueReconciling()` when true - it reduces the necessary indentation and doesn't signal a requeue when no changes to cluster state have been committed.

That said, is this Route supposed to always exist, or are there situations when it won't? I was originally under the impression that this would only be here when mTLS was actually in use.

rwhundley (changes requested):

controllers/operator/routes.go
Tirumalavasa: @rwhundley , addressed your review comment. 
On the point 2, your thought is right. but currently it is always exist. As We don't know when Customer registers an identity provider configuration for certificate based authentication, Operator doesn't know unless it contacts the DB and check for this in the DB and it also doesn't know when to contact DB to see if cert based authentication is enabled or disabled. To avoid overhead of making Db calls we are creating this route by default.

controllers/operator/routes.go
```diff
@@ -331,7 +331,10 @@ func (r *AuthenticationReconciler) reconcileRoute(authCR *operatorv1alpha1.Authe
 
 func (r *AuthenticationReconciler) ensureRouteDoesNotExist(ctx context.Context, authCR *operatorv1alpha1.Authentication, fields *reconcileRouteFields) (result *ctrl.Result, err error) {
 	reqLogger := logf.FromContext(ctx)
-	if fields.Name != IMCrtAuthRouteName {
+	if fields.Name == IMCrtAuthRouteName {
+		//do not delete certAuth route
+		return subreconciler.ContinueReconciling()
+	} else {
```
rwhundley, line 337: You don't need this `else` because we're only dealing with the non-mTLS Routes for the rest of the function.

Tirumalavasa: removed else block

rwhundley (approved):

PR #985, opened 2025-03-18
update master host to add master_host environment value from configmap

rashmi43:
Add MASTER_HOST environment variable to all the platform-auth containers after reading it from platform-auth-idp configmap


Tirumalavasa (approved):
/lgtm

PR #995, opened 2025-03-19
[SC2] - update master host using platform-auth-idp configmap

rashmi43:
Update MASTER_HOST from platform-auth-idp configmap rather than updating it in the deployment

https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65746


controllers/operator/configmap.go
```diff
@@ -265,6 +267,12 @@ func (r *AuthenticationReconciler) handleConfigMap(instance *operatorv1alpha1.Au
 					currentConfigMap.Data["PREFERRED_LOGIN"] = newConfigMap.Data["PREFERRED_LOGIN"]
 					cmUpdateRequired = true
 				}
+				if _, keyExists := currentConfigMap.Data["MASTER_HOST"]; !keyExists {
```
rwhundley, line 270: Note that if you have the old ClusterCADomain value set from previous version, an upgrade will not replace this with the cluster_address given this check - the key exists in that case.

rwhundley (changes requested):

controllers/operator/configmap.go
rwhundley:
```suggestion
				newConfigMap = functionList[index](instance, r.Scheme)
				if currentMasterHost, keyExists := currentConfigMap.Data["MASTER_HOST"]; !keyExists || currentMasterHost != newConfigMap.Data["MASTER_HOST"] {
```

rashmi43: checking for keyExists

controllers/operator/configmap.go
```diff
@@ -929,6 +941,20 @@ func isPublicCloud(client client.Client, namespace string, configMap string) boo
 	return false
 }
 
+// Check if hosted on IBM Cloud
+func getClusterAddress(client client.Client, namespace string, configMap string) string {
```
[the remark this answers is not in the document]
rashmi43: returned error as well, but not doing anything for error condition

controllers/operator/configmap.go
```diff
@@ -504,6 +512,10 @@ type registrationJSONData struct {
 func (r *AuthenticationReconciler) authIdpConfigMap(instance *operatorv1alpha1.Authentication, scheme *runtime.Scheme) *corev1.ConfigMap {
 	reqLogger := log.WithValues("Instance.Namespace", instance.Namespace, "Instance.Name", instance.Name)
 	isPublicCloud := isPublicCloud(r.Client, instance.Namespace, "ibmcloud-cluster-info")
+	clusterAddress := getClusterAddress(r.Client, instance.Namespace, "ibmcloud-cluster-info")
+	if strings.EqualFold(clusterAddress, "") {
```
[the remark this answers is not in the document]
rashmi43: addressed, wondering if I should check this or error condition

controllers/operator/configmap.go
```diff
@@ -22,6 +22,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
+	myerr "errors"
```
rwhundley, line 25: Would you please instead avoid naming this import and name the `k8s.io/apimachinery/pkg/api/errors` package "k8sErrors" - this is what I've been doing in other files whenever something like this comes up.

controllers/operator/configmap.go
```diff
@@ -267,7 +268,7 @@ func (r *AuthenticationReconciler) handleConfigMap(instance *operatorv1alpha1.Au
 					currentConfigMap.Data["PREFERRED_LOGIN"] = newConfigMap.Data["PREFERRED_LOGIN"]
 					cmUpdateRequired = true
 				}
-				if _, keyExists := currentConfigMap.Data["MASTER_HOST"]; !keyExists {
+				if _, keyExists := currentConfigMap.Data["MASTER_HOST"]; keyExists {
```
rwhundley, line 271: This will always overwrite the value and trigger an update on the ConfigMap; I was thinking that you would instead compare the value that you retrieve to the generated one, compare the two, and only set/update when they differ.

controllers/operator/configmap.go
```diff
@@ -512,8 +513,12 @@ type registrationJSONData struct {
 func (r *AuthenticationReconciler) authIdpConfigMap(instance *operatorv1alpha1.Authentication, scheme *runtime.Scheme) *corev1.ConfigMap {
 	reqLogger := log.WithValues("Instance.Namespace", instance.Namespace, "Instance.Name", instance.Name)
 	isPublicCloud := isPublicCloud(r.Client, instance.Namespace, "ibmcloud-cluster-info")
-	clusterAddress := getClusterAddress(r.Client, instance.Namespace, "ibmcloud-cluster-info")
-	if strings.EqualFold(clusterAddress, "") {
+	clusterAddress, cerr := getClusterAddress(r.Client, instance.Namespace, "ibmcloud-cluster-info")
+	if cerr != nil {
+		reqLogger.Error(cerr, "Failed to fetch the cluster_address from the ibmcloud-cluster-info ConfigMap")
+		return nil
```
rwhundley, line 519: If this prompts a return of nil, that means that we need to check for nil in all of the cases where we are accessing fields on the return value of this ConfigMap. This is an oversight that's present throughout the handleConfigMap function in that update portion where the currentConfigMap is having its values set to the newConfigMap ones.

rwhundley (changes requested):

controllers/operator/configmap.go
rashmi43: sure

controllers/operator/configmap.go
```diff
@@ -238,15 +241,16 @@ func (r *AuthenticationReconciler) handleConfigMap(instance *operatorv1alpha1.Au
 			// @posriniv - find a more efficient solution
 			if configMapList[index] == "platform-auth-idp" {
 				cmUpdateRequired := false
+				if newConfigMap, err = functionList[index](instance, r.Scheme); err != nil {
```
rashmi43, line 244: we never returned an error here, would this case any new issues @rwhundley

PR #1003, opened 2025-03-25
Get from API server when cache fails

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#66250](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/66250)

Fall back to querying the API server directly when an object isn't found just in case the cache filtering is leading to missing existing objects.


Tirumalavasa (approved):
/lgtm

PR #1015, opened 2025-04-09
Ford Hotfix GH #65990

rwhundley:


controllers/operator/configmap.go
```diff
@@ -272,6 +272,11 @@ func (r *AuthenticationReconciler) handleConfigMap(instance *operatorv1alpha1.Au
 					currentConfigMap.Data["MASTER_HOST"] = newConfigMap.Data["MASTER_HOST"]
 					cmUpdateRequired = true
 				}
+				if _, keyExists := currentConfigMap.Data["MASTER_PATH"]; !keyExists {
```
rwhundley, line 275: The stanza immediately preceding this one is dealing with MASTER_PATH - why is this one necessary?

controllers/operator/configmap.go
```diff
@@ -187,6 +187,7 @@ func (r *AuthenticationReconciler) handleConfigMap(instance *operatorv1alpha1.Au
 						return err
 					}
 					if configMapList[index] == "platform-auth-idp" {
+						newConfigMap.Data["MASTER_PATH"] = "/idauth"
```
rwhundley: If this is always being set, this should just be set directly in the `authIdpConfigMap` function.

PR #1031, opened 2025-05-13
Expose Routes externally with front door

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#66643](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/66643)


controllers/operator/routes.go
```diff
@@ -588,6 +588,9 @@ func (r *AuthenticationReconciler) getClusterAddress(authCR *operatorv1alpha1.Au
 		clusterInfoConfigMap := &corev1.ConfigMap{}
 
 		clusterAddressFieldName := "cluster_address"
+		if authCR.Spec.Config.ZenFrontDoor && ctrlcommon.ClusterHasZenExtensionGroupVersion(&r.DiscoveryClient) {
```
ujjwalchk-it, line 591: Any reason of using `cluster_address_auth` instead of `cluster_address` for ZenFrontDoor ?

rwhundley: `cluster_address_auth` has essentially meant "the host that is delegated to us from elsewhere" since it was introduced for ZenExtension reasons. Then we made them always the same value, so maybe it matters less. And, in an ideal world, I'd get rid of all of these redundant values and have less to track. Unfortunately, I have no idea who or what is looking at which fields when, so this is being done mostly as a hedge against that.

rwhundley: I suppose the next question is why am I not using the function that's defined 150 lines earlier that is a more readable shorthand for the imperative conditional here? Probably because I was tired. 🙃 Given this has passed internal validations, I'll go ahead and start cleaning this branch up.

controllers/operator/configmap.go
```diff
@@ -834,7 +834,7 @@ func (r *AuthenticationReconciler) generateCNCFClusterInfo(ctx context.Context,
 	clusterAddress := strings.Join([]string{strings.Join([]string{"cp-console", authCR.Namespace}, "-"), domainName}, ".")
 	clusterEndpoint := "https://" + clusterAddress
 	clusterAddressAuth := clusterAddress
-	if authCR.Spec.Config.ZenFrontDoor && ctrlcommon.ClusterHasZenExtensionGroupVersion(&r.DiscoveryClient) {
+	if shouldUseCPDHost(authCR, &r.DiscoveryClient) {
```
rwhundley, line 837: I just happened to notice this could be replaced by the function I had defined in routes.go, so I went ahead and did that so that it's being reused where appropriate.

rwhundley:
Putting this on hold until next release.

Tirumalavasa (approved):
/lgtm

PR #1033, opened 2025-05-14
Protect platform-auth-idp from upgrade clobber

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#66581](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/66581)

### DIFF
--- a/controllers/bootstrap/authentication_bootstrap_controller.go
+++ b/controllers/bootstrap/authentication_bootstrap_controller.go
@@ -1,0 +1,176 @@
+/*
+Copyright 2025 IBM Corporation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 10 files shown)


Tirumalavasa (approved):

controllers/bootstrap/authentication_bootstrap_controller.go
```diff
@@ -0,0 +1,172 @@
...
+// writeConfigurationsToAuthenticationCR copies values from the
+// platform-auth-idp ConfigMap to the Authentication CR.
+func (r *BootstrapReconciler) writeConfigurationsToAuthenticationCR(ctx context.Context, authCR *operatorv1alpha1.Authentication) (err error) {
+	platformAuthIDPCM := &corev1.ConfigMap{}
+	if err = r.Get(ctx, types.NamespacedName{Name: "platform-auth-idp", Namespace: authCR.Namespace}, platformAuthIDPCM); k8sErrors.IsNotFound(err) {
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("failed to get ConfigMap: %w", err)
+	}
+	keys := map[string]any{
```
ujjwalchk-it, line 114: Are these only keys which can be customisable ?

ujjwalchk-it:
In general - I have another question : would there be any race condition between when we update the AuthCR and being updated by any other process same time ?

controllers/bootstrap/authentication_bootstrap_controller.go
rwhundley: Good point - I've added the missing keys to both this list as well as to the list that updatesAlways runs on.

PR #1035, opened 2025-05-15
Make bundle at end of update-version

rwhundley:
When running `make update-version`, the bundles are not recreated, meaning that the changes that should be propagated to the bundle are not.


Tirumalavasa (approved):
/lgtm

PR #1060, opened 2025-07-02
[67117] Changes to connect to EDB using password

abhishekp03:
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/67117

### DIFF
--- a/internal/database/connectors/dbconn.go
+++ b/internal/database/connectors/dbconn.go
@@ -66,7 +66,12 @@ func NewPostgresDB(opts ...DBOption) (*PostgresDB, error) {
 }
 
 func (p *PostgresDB) Connect(ctx context.Context) (err error) {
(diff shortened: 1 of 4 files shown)


internal/database/connectors/dbconn.go
```diff
@@ -66,7 +66,12 @@ func NewPostgresDB(opts ...DBOption) (*PostgresDB, error) {
 }
 
 func (p *PostgresDB) Connect(ctx context.Context) (err error) {
-	dsn := fmt.Sprintf("host=%s user=%s dbname=%s port=%s sslmode=require", p.Host, p.User, p.Name, p.Port)
+	var dsn string
+	if len(p.Password) > 0 {
```
Tirumalavasa, line 70: @abhishekp03 , it is safer to add not nil check here `if p.Password != nil && len(*p.Password) > 0`

PR #1062, opened 2025-07-02
[SC2] - Update audit tls to deployments

rashmi43:


controllers/operator/deployment.go
```diff
@@ -1021,4 +1036,47 @@ func buildIdpVolumes(ldapCACert string, routerCertSecret string) []corev1.Volume
...
+func EnsureVolumePresent(volumes []corev1.Volume, newVol corev1.Volume) []corev1.Volume {
+	for _, v := range volumes {
+		if v.Name == newVol.Name {
+			return volumes // already exists
+		}
+	}
+	return append(volumes, newVol)
+}
+
+func IMAuditTLSVolume() corev1.Volume {
```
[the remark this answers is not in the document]
rashmi43: I can declare it as a variable

controllers/common/constants.go
```diff
@@ -34,6 +34,10 @@ const MongoStatefulsetName string = "icp-mongodb"
 // Name of CommonService created by IM Operator to provision EDB share
 const DatastoreEDBCSName string = "im-common-service"
 
+// Name of Secret containing certificates for Common Audit Logging
+const AuditTLSSecretName string = "audit-tls"
```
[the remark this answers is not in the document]
rashmi43: ok

controllers/operator/deployment.go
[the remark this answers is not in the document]
rashmi43: addressed

controllers/operator/configmap.go
```diff
@@ -534,6 +544,8 @@ func (r *AuthenticationReconciler) authIdpConfigMap(instance *operatorv1alpha1.A
 	// Set the path for SAML connections
 	masterPath := "/idauth"
 
+	auditURL, auditSecret := getAuditEndpointDetails(r.Client, instance.Namespace, "audit-endpoint")
```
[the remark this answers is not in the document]
rashmi43: rob said we want to stop using this file, in his earlier review comments

controllers/operator/deployment.go
```diff
@@ -1021,4 +1036,47 @@ func buildIdpVolumes(ldapCACert string, routerCertSecret string) []corev1.Volume
...
+
+	if auditSecretExists && required {
+		volumes = EnsureVolumePresent(volumes, IMAuditTLSVolume())
+	}
+	return volumes
+}
+
+// EnsureVolumePresent checks if a volume exists
+// If not, it appends the new volume and returns the updated slice.
+func EnsureVolumePresent(volumes []corev1.Volume, newVol corev1.Volume) []corev1.Volume {
```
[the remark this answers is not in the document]
rashmi43: Addressed

rashmi43:
```
sh-5.1$ env|grep AUDIT
AUDIT_URL=https://common-audit-logging.cp4s.svc.cluster.local:9880
AUDIT_ENABLED=true

sh-5.1$

File: controllers/operator/configmap.go [hunk 5]
```diff
@@ -966,6 +974,46 @@ func getClusterAddress(client client.Client, namespace string, configMap string)
...
+	if err := client.Get(context.TODO(), types.NamespacedName{Name: auditConfigMapName, Namespace: namespace}, auditConfigMap); err != nil {
+		if k8sErrors.IsNotFound(err) {
+			log.Error(err, "The configmap ", auditConfigMapName, " is not created yet")
+			// no requeue required
+		}
+		log.Error(err, "Failed to get Audit endpoint configmap", "Secret.Name", auditConfigMapName)
+		return ""
+	}
+	auditURL = auditConfigMap.Data["audit-url"]
+	auditSecretName = auditConfigMap.Data["audit-secret"]
```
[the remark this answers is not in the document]
rashmi43: we are not saving it in platform-auth-idp configmap, but we read it to check its existencewe are not saving it in platform-auth-idp configmap, but we read it to check its existence

rwhundley (changes requested):

controllers/operator/configmap.go
```diff
@@ -966,6 +974,46 @@ func getClusterAddress(client client.Client, namespace string, configMap string)
...
+	if len(auditSecretName) > 0 && CheckSecretExists(client, namespace, auditSecretName) {
+		return auditURL
+	}
+	return ""
+}
+
+func CheckSecretExists(client client.Client, namespace string, auditSecretName string) bool {
+	auditTLSSecret := &corev1.Secret{}
+	auditTLSSecretStruct := types.NamespacedName{Name: auditSecretName, Namespace: namespace}
+	if err := client.Get(context.TODO(), auditTLSSecretStruct, auditTLSSecret); err != nil {
```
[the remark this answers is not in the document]
rashmi43: addressed

controllers/operator/configmap.go
```diff
@@ -966,6 +976,18 @@ func getClusterAddress(client client.Client, namespace string, configMap string)
 	return "", errors.New(fmt.Sprint("failed to fetch the cluster address"))
 }
 
+func CheckSecretExists(client client.Client, namespace string, auditSecretName string) (bool, error) {
```
rwhundley, line 979: Probably rename this to `checkSecretExists` given we don't need to expose this function outside of this package.

controllers/operator/configmap.go
```diff
@@ -268,6 +268,16 @@ func (r *AuthenticationReconciler) handleConfigMap(instance *operatorv1alpha1.Au
...
+				if _, keyExists := currentConfigMap.Data["AUDIT_URL"]; !keyExists {
+					reqLogger.Info("Updating an existing Configmap", "Configmap.Namespace", currentConfigMap.Namespace, "ConfigMap.Name", currentConfigMap.Name)
+					currentConfigMap.Data["AUDIT_URL"] = newConfigMap.Data["AUDIT_URL"]
+					cmUpdateRequired = true
+				}
+				if _, keyExists := currentConfigMap.Data["AUDIT_SECRET"]; !keyExists {
+					reqLogger.Info("Updating an existing Configmap", "Configmap.Namespace", currentConfigMap.Namespace, "ConfigMap.Name", currentConfigMap.Name)
+					currentConfigMap.Data["AUDIT_SECRET"] = newConfigMap.Data["AUDIT_SECRET"]
+					cmUpdateRequired = true
+				}
```
rwhundley, line 280: You can get rid of these - if the kvps are in the ConfigMap, we want to keep them as-is. If we had fields on the Authentication CR or some set of defaults we wanted to set, we would do that here. But not the case for this feature.

rashmi43: removing it, do we need this when we upgrade and the CM exists?

controllers/operator/deployment.go
```diff
@@ -822,8 +870,31 @@ func generateManagerDeploymentObject(instance *operatorv1alpha1.Authentication,
...
 }
 
-func buildIdpVolumes(ldapCACert string, routerCertSecret string) []corev1.Volume {
-	return []corev1.Volume{
+func buildIdpVolumes(ldapCACert string, routerCertSecret string, auditSecretExists bool, required bool) []corev1.Volume {
+	auditVolume := corev1.Volume{
+		Name: IMAuditTLSVolume,
+		VolumeSource: corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName: AuditTLSSecretName,
```
rwhundley, line 878: This isn't always the Secret name anymore - the name's being defined in `platform-auth-idp` now.

rwhundley (changes requested):

controllers/operator/deployment.go
rashmi43: fixed

controllers/operator/deployment.go
```diff
@@ -480,7 +525,7 @@ func generateDeploymentObject(instance *operatorv1alpha1.Authentication, scheme
 							Operator: corev1.TolerationOpExists,
 						},
 					},
-					Volumes:        buildIdpVolumes(ldapCACert, routerCertSecret),
+					Volumes:        buildIdpVolumes(ldapCACert, routerCertSecret, false, false),
```
rwhundley: This doesn't have enough arguments.

controllers/operator/deployment.go
```diff
@@ -822,8 +867,31 @@ func generateManagerDeploymentObject(instance *operatorv1alpha1.Authentication,
 	return idpDeployment
 }
 
-func buildIdpVolumes(ldapCACert string, routerCertSecret string) []corev1.Volume {
-	return []corev1.Volume{
+func buildIdpVolumes(ldapCACert string, routerCertSecret string, auditSecretExists bool, required bool, auditSecretName string) []corev1.Volume {
```
rwhundley, line 870: What is `required` for? Also, providing a Secret name for the TLS Secret would seem to imply that the volume for the audit TLS Secret is needed, so you could just submit the `auditSecretName` and add the volume if it is a non-empty string. Would also require a change ahead of calling this function to be sure that `auditSecretName` is set to `""` if the Secret isn't present. Or you could just get the audit Secret within this code. Whatever you want to do.

rwhundley (changes requested):

controllers/operator/deployment.go
rashmi43: its fixed, if you refresh

controllers/operator/deployment.go
rashmi43: provider and management need the audit volume, but auth-service does not, I am passing a false for auth-service and true for the other two

controllers/operator/deployment.go
```diff
@@ -106,10 +109,31 @@ func (r *AuthenticationReconciler) handleDeployment(instance *operatorv1alpha1.A
 		saasServiceIdCrn = saasTenantConfigMap.Data["service_crn_id"]
 	}
 
+	var auditSecretExists bool
```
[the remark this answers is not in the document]
rashmi43: ok

controllers/operator/deployment.go
rwhundley: The way that you could achieve the same effect with the signature I'm suggesting is as follows:

- provider and management: `buildIdpVolumes(ldapCACert, routerCertSecret, auditSecretName)`
- auth service: `buildIdpVolumes(ldapCACert, routerCertSecret, "")`

controllers/operator/deployment.go
```diff
@@ -659,7 +704,7 @@ func generateProviderDeploymentObject(instance *operatorv1alpha1.Authentication,
 	return idpDeployment
 }
 
-func generateManagerDeploymentObject(instance *operatorv1alpha1.Authentication, scheme *runtime.Scheme, deployment string, icpConsoleURL string, saasCrnId string) *appsv1.Deployment {
+func generateManagerDeploymentObject(instance *operatorv1alpha1.Authentication, scheme *runtime.Scheme, deployment string, icpConsoleURL string, saasCrnId string, auditSecretExists bool, auditSecretName string) *appsv1.Deployment {
```
rwhundley, line 707: What I've said RE: using the `auditSecretName` as the signal for whether the Secret exists applies here, too - you should be able to just check to see if the name is empty and then request the audit tls volumes when it isn't.

rwhundley (changes requested):
There isn't any extra information conveyed by having a boolean alongside the audit secret name in the contexts I've commented on - the code can readily be structured to treat a non-empty string name value as an indication that there are volumes/volume mounts needed for the audit TLS Secret, and this will make the function signatures more concise/clear.

controllers/operator/deployment.go
```diff
@@ -33,6 +33,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+// Name of Secret containing certificates for Common Audit Logging
+const AuditTLSSecretName string = "audit-tls"
```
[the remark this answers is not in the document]
rashmi43: removed

controllers/operator/deployment.go
```diff
@@ -106,43 +106,16 @@ func (r *AuthenticationReconciler) handleDeployment(instance *operatorv1alpha1.A
...
-		if len(auditSecretName) == 0 || len(auditURL) == 0 {
-			reqLogger.Info("Unable to fetch the audit url and audit secret from auth-idp configmap", authIdpConfigMapName)
-		} else if len(auditSecretName) > 0 {
-			auditSecretExists, err = checkSecretExists(r.Client, instance.Namespace, auditSecretName)
-			if err != nil {
-				return err
-			}
-			reqLogger.Info("audit-tls secret name", "Deployment.Namespace", instance.Namespace, "Secret.Name", auditSecretName)
-		}
+	auditSecretName, err := fetchAuditDetails(r.Client, instance.Namespace)
```
rwhundley, line 109:
```suggestion
	auditSecretName, err := r.getAuditSecretNameIfExists(context.TODO(), instance.Namespace)
```

controllers/operator/deployment.go
```diff
@@ -353,6 +326,40 @@ func (r *AuthenticationReconciler) handleDeployment(instance *operatorv1alpha1.A
...
+		} else if len(auditSecretName) > 0 {
+			_, err2 := checkSecretExists(client, namespace, auditSecretName)
+			if err2 != nil {
+				return "", err2
+			}
+			log.Info("audit-tls secret name", "Deployment.Namespace", namespace, "Secret.Name", auditSecretName)
+		}
+	}
+	return auditSecretName, nil
+}
```
rwhundley, line 361:
```suggestion
// getAuditSecretNameIfExists determines whether an audit service has been
// configured with TLS. Returns the name of the Secret used to store the TLS
// certificates if a Secret has been identified by the user and found on the
// cluster, or an empty string when the Secret isn't found or cannot otherwise
// be retrieved. If an error other than NotFound is received when trying to get
// the Secret, that is returned as well.
func (r *AuthenticationReconciler) getAuditSecretNameIfExists(ctx context.Context, namespace string) (string, error) {
	var auditSecretName string
	var auditURL string
	// Check for the presence of audit-endpoint configmap
	authIdpConfigMapName := "platform-auth-idp"
	authIdpConfigMap := &corev1.ConfigMap{}
	idpCMLogger := log.WithValues("ConfigMap.Name", authIdpConfigMapName, "ConfigMap.Namespace", namespace)
	if err := r.Get(ctx, types.NamespacedName{Name: authIdpConfigMapName, Namespace: namespace}, authIdpConfigMap); errors.IsNotFound(err) {
		idpCMLogger.Info("ConfigMap was not found")
		return "", nil
	} else if err != nil {
		idpCMLogger.Error(err, "Failed to get ConfigMap")
		return "", err
	}
	if authIdpConfigMap.Data == nil {
		idpCMLogger.Info("Invalid ConfigMap")
		return "", nil
	}
	if authIdpConfigMap.Data["AUDIT_URL"] == "" {
		idpCMLogger.Info("Audit URL is not specified in ConfigMap; assume no Secret to mount", "key", "AUDIT_URL")
		return "", nil
	}
	if authIdpConfigMap.Data["AUDIT_SECRET"] == "" {
		idpCMLogger.Info("Audit Secret is not specified in ConfigMap; assume no Secret", "key", "AUDIT_SECRET")
		return "", nil
	}
	auditURL = authIdpConfigMap.Data["AUDIT_URL"]
	auditSecretName = authIdpConfigMap.Data["AUDIT_SECRET"]
	idpCMLogger.Info("Fetched audit URL and audit Secret from ConfigMap", "AUDIT_SECRET", auditSecretName, "AUDIT_URL", auditURL)

	auditTLSSecretLogger := log.WithValues("Secret.Name", auditSecretName, "Secret.Namespace", namespace)
	auditTLSSecret := &corev1.Secret{}
	auditTLSSecretStruct := types.NamespacedName{Name: auditSecretName, Namespace: namespace}
	err := r.Get(ctx, auditTLSSecretStruct, auditTLSSecret)
	if errors.IsNotFound(err) {
		auditTLSSecretLogger.Info("Secret for audit configuration not found")
		return "", nil
	} else if err != nil {
		auditTLSSecretLogger.Error(err, "Failed to retrieve Secret for audit configuration")
		return "", err
	}

	auditTLSSecretLogger.Info("Secret found for audit configuration")
	return auditSecretName, nil
}
```

rwhundley (changes requested):

rwhundley (approved):
Alright, seems good.

Tirumalavasa (approved):

PR #1066, opened 2025-07-15
Add flag to enable extra attribute for JIT

YangShyrMei:
Issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/66941
Requirement : Additional attribute like "email" , "username" should be propagated to the JIT group members section for the CPD to display this .

Design discussion:
Use a similar approach of having a flag like the ATTR_MAPPING_FROM_CONFIG in platform-auth-idp cm which when enabled would send the additional attributes - email and username fro JIT Groups. This would ensure that the standards for JIT is maintained and would not cause any backward compatibility issues.

Therefore, in order to have these attributes --> the customer would need to enable the flag in the cm. Once enabled, the additional fields would be sent in the GET SCIM Group response allowing the fields to be populated in the UI.


YangShyrMei:
1. **ENABLE_JIT_EXTRA_ATTR** is populated with a default value - `false`
image

2. Updated it to `true`
image

3. The env variable - ENABLE_JIT_EXTRA_ATTR is available in platform-identity-mgmt pod
image

rwhundley (approved):
/lgtm

PR #1086, opened 2025-08-19
Perform migrations via Job

rwhundley:
Originating issues: [IBMPrivateCloud/roadmap#67379](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/67379), [IBMPrivateCloud/roadmap#67424](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/67424)

In order to work around limitations imposed by shipping OLM bundles, migration behaviors previously handled via a goroutine in the Authentication controller are now instead by a dedicated Job.

Additionally, the check performed for whether to use the `idauth` Route for SAML flows has also been broken out into a Job that is run in the event that the `MASTER_PATH` variable has not been set in the `platform-auth-idp` ConfigMap.

### DIFF
--- a/bundle/manifests/ibm-iam-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ibm-iam-operator.clusterserviceversion.yaml
@@ -57,6 +57,8 @@ metadata:
               }
             },
             "config": {
+              "auditSecret": "",
+              "auditUrl": "",
               "authUniqueHosts": "internal-ip1 internal-ip2 mycluster.icp",
               "bootstrapUserId": "kubeadmin",
               "claimsMap": "name=\"givenName\" family_name=\"givenName\" given_name=\"givenName\" preferred_username=\"displayName\" display_name=\"displayName\"",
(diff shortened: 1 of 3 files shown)


api/operator/v1alpha1/authentication_types.go
```diff
@@ -238,6 +230,8 @@ func NewMigrationInProgressCondition() *metav1.Condition {
 	}
 }
 
+// Creates a new ConditionMigrationsRunning condition for when the migration Job
+// has stopped running.
 func NewMigrationFinishedCondition() *metav1.Condition {
 	return &metav1.Condition{
 		Type:    ConditionMigrationsRunning,
```
Tirumalavasa, line 237: Hi @rwhundley , you meant  `ConditionMigrationsFinished` here?

rwhundley: The reason this distinction was made was because, as a single condition, the window of time where one could see that a migration had failed was very small before it would be updated to state that migrations were in progress. Tracking as two separate conditions allows for it to be possible to see from the Authentication CR that migrations have been failing repeatedly and whether or not migrations are actively running or not without as much of a timing constraint.

PR #1092, opened 2025-09-09
CD - liberty's webAppSecurity sameSiteCookie config addition

Tirumalavasa:
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/67624


rwhundley (commented):
This seems like something that should be managed through the Authentication CR.

internal/controller/operator/configmap.go
```diff
@@ -374,6 +375,8 @@ func updatePlatformAuthIDP(_ common.SecondaryReconciler, _ context.Context, obse
 			"LDAP_CTX_POOL_PREFERREDSIZE"),
 		updatesValuesWhen(not(observedKeySet[*corev1.ConfigMap]("MASTER_PATH")),
 			"MASTER_PATH"),
+		updatesValuesWhen(not(observedKeySet[*corev1.ConfigMap]("LIBERTY_SAMESITE_COOKIE")),
+			"LIBERTY_SAMESITE_COOKIE"),
```
rwhundley, line 379: We shouldn't need these lines because the value is already registered to always be updated on R327; you should remove this.

rwhundley (changes requested):
The other thing I'm looking for is adding LIBERTY_SAMESITE_COOKIE to the bootstrap controller's list of keys to preserve in platform-auth-idp. You should just need to register it in the `keys` map like AUDIT_SECRET given that's also an optional string.

PR #1101, opened 2025-09-17
Authsvc secret vol update

ujjwalchk-it:
Populating the default-admin and scim-admin credentials from the volume mount rather than the env var.
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/67843


rwhundley (commented):
Looks fine; just need to see the auth service changes get approved and then I can do the same here.

Tirumalavasa (approved):

PR #1108, opened 2025-09-18
Logging overhaul, part 2

rwhundley:
* Further divided logging to be more terse by default
* Fixed an issue where wait for Cluster CR would block when configured to use external EDB


sgrube (approved):
/lgtm

PR #1259, opened 2026-05-01
Remove Client CR update for custom hostname change

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#69306](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69306)

### DIFF
--- a/internal/controller/operator/authentication_controller.go
+++ b/internal/controller/operator/authentication_controller.go
@@ -337,7 +337,6 @@ func (r *AuthenticationReconciler) runNonStatusSubreconcilers(ctx context.Contex
 		r.handleUIOperandRequest,
 		r.handleRoutes,
 		r.handleHPAs,
-		r.syncClientHostnames,
 		r.handleMongoDBCleanup,
 		r.cleanupOldRBAC)...).Reconcile(ctx)
 }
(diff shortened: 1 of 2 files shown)


internal/controller/operator/authentication_controller.go
```diff
@@ -337,7 +337,6 @@ func (r *AuthenticationReconciler) runNonStatusSubreconcilers(ctx context.Contex
 		r.handleUIOperandRequest,
 		r.handleRoutes,
 		r.handleHPAs,
-		r.syncClientHostnames,
```
ujjwalchk-it, line 340: Are you intended to keep the main function still ?

```
func (r *AuthenticationReconciler) syncClientHostnames(ctx context.Context, req ctrl.Request) (result *ctrl.Result, err…
```

Tirumalavasa (dismissed):

internal/controller/operator/authentication_controller.go
rwhundley: I'm keeping the function for the time being because I may reuse a chunk of it for another request we have. If that doesn't come to pass for that issue, I will remove this function as well.

Tirumalavasa (approved):

PR #1261, opened 2026-05-06
increase liberty logs volumesize

Tirumalavasa:
Addresses : https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69127

### DIFF
--- a/bundle/manifests/ibm-iam-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/ibm-iam-operator.clusterserviceversion.yaml
@@ -29,7 +29,7 @@ metadata:
               "resources": {
                 "limits": {
                   "cpu": "1000m",
-                  "ephemeral-storage": "600Mi",
+                  "ephemeral-storage": "1500Mi",
                   "memory": "1Gi"
                 },
                 "requests": {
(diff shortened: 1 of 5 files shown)


rwhundley (commented):
I thought we were going to tie the increase in ephemeral/emptyDir limits to whether Liberty trace was enabled given under normal circumstances we aren't going to need that kind of space.

Tirumalavasa:
> I thought we were going to tie the increase in ephemeral/emptyDir limits to whether Liberty trace was enabled given under normal circumstances we aren't going to need that kind of space.

yes, But one can enable traces by dropping tracespec.xml into the configdroppins/ovverides. In this case ibm-iam-operator will not have knowledge on whether the trace is enabled or not as this is not configmap driven approach and also as ephemeral volume is within the 2GB limit, at least for this release we are going with this value and we can try to improve it in upcoming releases if any

rwhundley (approved):

rwhundley:
For the record, we don't want anyone doing the thing that you've described as a persistent strategy, and we do want customers who are enabling trace to do so via the ConfigMap as that will reduce the effort required to re-enable the trace reliably. It's the same idea as not encouraging customers to drop in server config overrides for adding/updating/removing IdPs or otherwise configuring Liberty - they could do that, but it will not persist reliably.

It doesn't make sense to keep the limit at such a high number when the vast majority of the time for the vast majority of installs, they should not need this kind of space, and there is a clear signal for when that kind of space is needed - enabling Liberty trace.

Tirumalavasa:
> For the record, we don't want anyone doing the thing that you've described as a persistent strategy, and we do want customers who are enabling trace to do so via the ConfigMap as that will reduce the effort required to re-enable the trace reliably. It's the same idea as not encouraging customers to drop in server config overrides for adding/updating/removing IdPs or otherwise configuring Liberty - they could do that, but it will not persist reliably.
> 
> It doesn't make sense to keep the limit at such a high number when the vast majority of the time for the vast majority of installs, they should not need this kind of space, and there is a clear signal for when that kind of space is needed - enabling Liberty trace.

Noted.

PR #1265, opened 2026-05-13
Account for OperandRequest health in readiness

rwhundley:
Originating issue: [IBMPrivateCloud/roadmap#69375](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69375)

### DIFF
--- a/internal/controller/operator/resourcestatus.go
+++ b/internal/controller/operator/resourcestatus.go
@@ -340,6 +340,54 @@ func getAllRouteStatus(ctx context.Context, k8sClient client.Client, names []str
 	return
 }
 
+func getOperandRequestStatus(ctx context.Context, k8sClient client.Client, namespacedName types.NamespacedName) (status operatorv1alpha1.ManagedResourceStatus) {
+	reqLogger := logf.FromContext(ctx).WithName("getOperandRequestStatus").V(1)
+	kind := "OperandRequest"
+	status = operatorv1alpha1.ManagedResourceStatus{
+		ObjectName: namespacedName.Name,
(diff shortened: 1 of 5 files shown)


internal/controller/operator/resourcestatus.go
```diff
@@ -368,6 +416,21 @@ func (r *AuthenticationReconciler) getCurrentServiceStatus(ctx context.Context,
 			names: []string{"oidc-client-registration", MigrationJobName},
 			f:     getAllJobStatus,
 		},
+		{
+			names: []string{"im-needs-ui", "im-needs-database"},
+			f:     getAllOperandRequestStatus,
+		},
+	}
+
+	opReqStatusRetrieval := statusRetrieval{
```
Tirumalavasa: The opReqStatusRetrieval variable is defined separately and then conditionally appended to statusRetrievals, but it's already included in the statusRetrievals slice initialization (lines 419-422 in the diff). This creates duplicate entries when ClusterHasOperandRequestAPIResource returns true, causing the same OperandRequests to be checked twice.

Tirumalavasa (changes requested):

Tirumalavasa (approved):

Tirumalavasa:
@rwhundley , did you get a chance to capture test results on this?

rwhundley:
@Tirumalavasa RE: testing: the unit tests in my view should be written to meet the burden of proof for integration. Review should highlight whether the testing has any relevant gaps that should be bridged. If something is found after that in integration testing, we need to question how thorough our review process is.

PR #1275, opened 2026-06-23
helm chart changes to support disabling route/ingress rbac permissions

Tirumalavasa:


Tirumalavasa:
Test results are captured here https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69628#issuecomment-213638143

cmd/main.go
```diff
@@ -77,26 +79,152 @@ func init() {
 		return
 	}
 
+	// Create a temporary client for SSAR checks during init
+	tempClient, err := client.New(cfg, client.Options{Scheme: scheme})
```
ujjwalchk-it, line 83: As we removing the roles for only routes and ingress API - any reason for adding the permission check for other APIs?

Tirumalavasa: @ujjwalchk-it ,Right. Rob has added SSAR checks to avoid panics in case of any missing api groups in the cluster

ujjwalchk-it (approved):

PR #1288, opened 2026-07-09
fix: unconditional scheme registration and retry SSAR checks at startup

Tirumalavasa:
When ibm-iam-operator is deployed in a namespace-scoped configuration (e.g. operator in cs-operators, Authentication CR in cs-data), ibm-namespace-scope- operator projects RBAC into the watched namespaces *after* the operator pod starts. PR #1275 introduced one-shot SSAR-gated scheme registration in init() and one-shot watch setup in SetupWithManager(), both of which race against that RBAC projection, causing two distinct failures:

1. 'no kind is registered for the type v1alpha1.OperandRequest in scheme' init() ran before RBAC was projected -> SSAR denied -> base scheme used (Authentication only, no OperandRequest/OperandBindInfo) -> scheme frozen for process lifetime -> every reconcile loop panics on OperandRequest Get.

2. 'Failed to watch' errors for HPA, Certificate, Service, etc. Cache informers try to list those types in cs-data before RBAC is ready.

Fixes:

cmd/main.go:
  Remove SSAR-gated branching from init(). All operator.ibm.com, Route,
  and Ingress types are now registered unconditionally. Scheme registration
  has no security implication; it only maps Go types to GVKs. Access control
  belongs in the watch setup, not the scheme. SecretProviderClass remains
  discovery-gated (requires CRD presence, not an RBAC race).

internal/controller/operator/authentication_controller.go:
  Add waitForAPIAccess() which retries hasAPIAccessInNamespaces every 2s
  for up to 2 minutes. All optional watches (Route, Ingress, OperandRequest,
  OperandBindInfo, SecretProviderClass) use this instead of the one-shot
  hasAPIAccessInNamespaces call. If ibm-namespace-scope-operator projects
  the RBAC within the 2-minute window, all watches are registered correctly
  without requiring a pod restart.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -29,7 +29,6 @@ import (
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
-	authorizationv1 "k8s.io/api/authorization/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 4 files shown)


Tirumalavasa:
closing this due to large diff

PR #1289, opened 2026-07-09
fix: unconditional ODLM scheme registration and SSAR retry for RBAC

Tirumalavasa:
Two bugs introduced by #1275 affect namespace-scoped deployments where ibm-namespace-scope-operator projects RBAC into watched namespaces after pod start (e.g. operator in cs-operators, Authentication CR in cs-data).

Bug 1 — "no kind is registered for the type v1alpha1.OperandRequest in scheme"
  init() in cmd/main.go ran a one-shot SSAR to decide which scheme to register.
  Because the SSAR fired before RBAC was projected, it returned denied, so
  operatorv1alpha1.AddToScheme (base scheme: Authentication only, no
  OperandRequest/OperandBindInfo) was used. The scheme is frozen for the
  process lifetime, so every reconcile call to r.Get(..., &OperandRequest{})
  panicked with "no kind registered". Deleting the pod worked only because
  on restart the RBAC was already in cs-data.

  Fix: always call operatorv1alpha1.AddODLMEnabledToScheme unconditionally.
  Scheme registration has no security implication — it only maps Go types to
  GVKs. Whether to *watch* a resource is gated in SetupWithManager.

Bug 2 — OperandRequest/OperandBindInfo watches permanently skipped
  SetupWithManager ran a one-shot SSAR for OperandRequest and OperandBindInfo.
  Same race: SSAR returned denied before RBAC was projected, so the watches
  were skipped for the lifetime of the process.

  Fix: replace the one-shot hasAPIAccessInNamespaces call for OperandRequest
  and OperandBindInfo with waitForODLMAccess, which retries every 2s for up
  to 5 minutes. Route, Ingress, and SPC checks are deliberately NOT retried:
  - Route: intentionally optional even on OCP (operator may not be configured to manage routes)
  - Ingress: intentionally optional on both CNCF and OCP
  - SPC: intentionally optional (requires CSI driver) A denied SSAR for those resources means "don't watch" by design, not a transient race. Retrying them would stall startup unnecessarily.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -99,25 +99,15 @@ func init() {
 		namespacesToCheck = strings.Split(watchNamespaceEnv, ",")
 	}
 
-	// Check OperandRequest permissions across all watch namespaces
-	operandRequestVerbs := []string{"create", "get", "list", "patch", "watch", "update", "delete"}
-	if controllercommon.ClusterHasOperandRequestAPIResource(dc) {
-		hasOperandRequestAccess, err := hasNamespacedAPIAccessForNamespaces(ctx, tempClient, namespacesToCheck, "operator.ibm.com", "operandrequests", operandRequestVerbs)
-		if err != nil {
-			setupLog.Error(err, "Failed to check OperandRequest permissions")
-			hasOperandRequestAccess = false
-		}
-		if hasOperandRequestAccess {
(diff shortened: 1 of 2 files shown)


Tirumalavasa:
Fixes: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69721

ujjwalchk-it (approved):

PR #1316, opened 2026-09-03
Introducing standardized statuses and operation timings

apxndy:
Fixes: [#69876](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69876)


apxndy:
Test result in comments on issue [starting here](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69876#issuecomment-250080348).

helm/values.yaml
```diff
@@ -17,6 +17,10 @@ global:
   imagePullPrefix: icr.io
   imagePullSecret: ibm-entitlement-key
   imagePullPolicy: IfNotPresent
+  # enforceLeastPrivilege removes the events RBAC permission and disables
+  # Kubernetes Event emission in the operator. Set to true for deployments
+  # that require a stricter security posture.
+  enforceLeastPrivilege: false
```
rwhundley: I don't like the name of this value for something that maps directly to whether or not the Events API specifically is being allowed or not. Whether something is or isn't "least privilege" should be evident by whether any/all gvr verbs beyond the minimum required are granted. It gives the impression that it is doing/controlling more than it actually is.

What should _actually_ happen is we expose the individual controls for what can be enabled, e.g. what we have for `manageNetworking` now, and the adopter consuming our software sets those fields appropriately to achieve whatever posture they want. If the adopter wants to have an `enforceLeastPrivilege` flag that abstracts away those concrete settings selections on our chart, that's fine and totally understandable.

internal/controller/operator/operationtiming.go
```diff
@@ -0,0 +1,250 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	operatorv1alpha1 "github.com/IBM/ibm-iam-operator/api/operator/v1alpha1"
+)
+
+const (
+	maxOperationTimingEntries = 5
```
rwhundley: Why 5? Why not 10? Or at least enough that it covers all of the different checkpoints?

internal/controller/operator/operationtiming.go
```diff
@@ -0,0 +1,250 @@
...
+
+// operation is an in-flight operation for one Authentication CR.
+type operation struct {
+	start metav1.Time
+	// deps are the external dependencies waited on, in wait order.
+	// A zero ReadyTime means the dependency is still pending.
+	deps []operatorv1alpha1.DependencyTime
+}
+
+func (o *operation) dep(component string) *operatorv1alpha1.DependencyTime {
```
rwhundley: The names of functions in this file are not the most readable - it would seem that this is a getter, so prefixing with `get` makes this read a lot more clearly.

internal/controller/operator/operationtiming.go
```diff
@@ -0,0 +1,250 @@
...
+			}
+			o.deps = slices.Delete(o.deps, i, i+1)
+			return true
+		}
+	}
+	return false
+}
+
+// pending reports whether any dependency waited on is not ready yet.
+func (o *operation) pending() bool {
```
rwhundley: Same here - if we're checking whether something is or isn't true as the primary mode of the function, prefixing with `is` makes that clearer.

internal/controller/operator/operationtiming.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	if t.ops == nil {
+		t.ops = make(map[types.NamespacedName]*operation)
+	}
+	op = &operation{start: start}
+	t.ops[key] = op
+	return op, true
+}
+
+func (t *operationTracker) remove(key types.NamespacedName) {
```
rwhundley: Some of these are missing doc comments; I think the comments would also help with the readability here.

internal/controller/operator/operationtiming.go
```diff
@@ -0,0 +1,250 @@
...
+		e.DependencyTime = slices.Clone(o.deps)
+	}
+	return e
+}
+
+// operationTracker holds in-flight operations keyed by CR. Kept in memory
+// across requeues so dependency waits span multiple passes. The mutex guards
+// the map only; controller-runtime serialises reconciles per CR.
+type operationTracker struct {
+	mu  sync.Mutex
```
rwhundley: What is this mutex guarding against exactly? Authentication is a singleton.

rwhundley: I believe mutexes only came into this project because of some slapdash implementation on my part for another incredibly similar change, but I don't know what utility they actually provide if we aren't kicking off goroutines that might clobber each other's changes to this data structure.

cmd/main.go
```diff
@@ -284,6 +285,16 @@ func main() {
 		os.Exit(1)
 	}
 
+	// With ENFORCE_LEAST_PRIVILEGE the events RBAC is not grated, so events are
+	// discarded rather that failing with Forbidden.
+	enforceLeastPrivilege := os.Getenv("ENFORCE_LEAST_PRIVILEGE") == "true"
```
rwhundley: We shouldn't need an env var for this - I'd instead execute an SSAR like we do for other objects where the authorizations aren't guaranteed. That will work regardless of whether it's installed via Helm or OLM.

internal/controller/operator/operationtiming.go
```diff
@@ -0,0 +1,250 @@
...
+		r.operations.remove(key)
+		logf.FromContext(ctx).Info("Recorded operationTiming", "totalDuration", entry.TotalDuration)
+		r.event(authCR, corev1.EventTypeNormal, EventReasonOperationEnded,
+			fmt.Sprintf("Operation completed for %s/%s in %s", authCR.Namespace, authCR.Name, entry.TotalDuration))
+	}
+}
+
+func (r *AuthenticationReconciler) event(authCR *operatorv1alpha1.Authentication, eventType, reason, message string) {
+	if r.Recorder != nil {
+		r.Recorder.Event(authCR, eventType, reason, message)
```
rwhundley: Isn't `r.Recorder` always getting set based upon what's in `main.go`?

internal/controller/operator/authentication_controller.go
```diff
@@ -625,28 +727,28 @@ func (r *AuthenticationReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-	//		predLog.Info("Delete event", "Label.Version", e.Object.GetLabels()[common.ManagerVersionLabel], "Controller.Version", version.Version, "match", e.Object.GetLabels()[common.ManagerVersionLabel] == version.Version)
-	//		return e.Object.GetLabels()[common.ManagerVersionLabel] == version.Version
-	//	},
-
-	//	// Allow generic events (e.g., external triggers)
-	//	GenericFunc: func(e event.GenericEvent) bool {
-	//		predLog.Info("Generic event", "Label.Version", e.Object.GetLabels()[common.ManagerVersionLabel], "Controller.Version", version.Version, "match", e.Object.GetLabels()[common.ManagerVersionLabel] == version.Version)
-	//		return e.Object.GetLabels()[common.ManagerVersionLabel] == version.Version
-	//	},
+	//  UpdateFunc: func(e event.UpdateEvent) bool {
```
rwhundley: Let's just remove this block of comments.

rwhundley: Or revert the change because it's noisy.

